PR #10, opened 2017-05-19
MULE-12256: don't require an extra slash if basePath is redundant

alepulver:


src/main/java/org/mule/extension/http/api/listener/server/HttpListenerConfig.java
```diff
@@ -55,6 +55,11 @@
   @Override
   public void initialise() throws InitialisationException {
     basePath = sanitizePathWithStartSlash(this.basePath);
+
+    // ignore base path if empty
+    if (basePath.equals("/")) {
```
afelisatti: Why can't this be part of sanitizePath...?

PR #25, opened 2017-06-05
MULE-12692: HTTP parsing should set correct mime type.

afelisatti:


src/main/java/org/mule/extension/http/internal/listener/HttpRequestToResult.java
```diff
@@ -104,6 +107,10 @@
     return Result.<Object, HttpRequestAttributes>builder().output(payload).mediaType(mediaType).attributes(attributes).build();
   }
 
+  private static MediaType getJavaMediaType(MediaType mediaType) {
+    return MediaType.create(APPLICATION_JAVA.getPrimaryType(), APPLICATION_JAVA.getSubType(), mediaType.getCharset().get());
```
estebanwasinger: This can be replaced with `APPLICATION_JAVA.charset(mediaType.getCharset().get()`

afelisatti: Thanks!

src/main/java/org/mule/extension/http/internal/listener/HttpResponseFactory.java
```diff
@@ -121,7 +122,8 @@ public HttpResponse create(HttpResponseBuilder responseBuilder,
       setupContentLengthEncoding(httpResponseHeaderBuilder, 0);
       httpEntity = new EmptyHttpEntity();
     } else if (payload instanceof Map) {
-      if (configuredContentType == null) {
+      if (configuredContentType == null || isJavaMimeType(configuredContentType)) {
```
elrodro83: any reason why `configuredContentType`is a String instead of a `MediaType`?

afelisatti: cause there's no actual MediaType involved here, only the String header value

PR #36, opened 2017-06-22
MULE-11606: Support a way to have DW handle multimaps

afelisatti:


src/main/java/org/mule/extension/http/internal/request/HttpRequestFactory.java
```diff
@@ -143,10 +143,10 @@ public HttpRequest create(HttpRequesterRequestBuilder requestBuilder, HttpAuthen
     return builder.build();
   }
 
-  private ParameterMap toParameterMap(Map<String, String> map) {
-    ParameterMap parameterMap = new ParameterMap();
-    map.forEach(parameterMap::put);
-    return parameterMap;
+  private MultiMap<String, String> toMultiMap(Map<String, String> map) {
```
elrodro83: doesn't a MultiMap constructor do this already?

afelisatti: Seems like it, though it also checks for collection values. In any case, this will be removed in another upcoming task which will turn the HTTP parameters into MultiMaps themselves.

PR #37, opened 2017-06-22
MULE-12866: HTTP Stream Response - remove switch

afelisatti:


src/main/java/org/mule/extension/http/internal/request/HttpRequesterProvider.java
```diff
@@ -106,6 +109,8 @@
 
   private TlsContextFactoryBuilder defaultTlsContextFactoryBuilder = TlsContextFactory.builder();
 
+  private boolean streamResponse = parseBoolean(getProperty(STREAM_RESPONSE_PROPERTY, "true"));
```
dfeist: Use `java.lang.Boolean#getBoolean`

afelisatti: Cleaner, but that default to false.

dfeist: Isn't it best to define the property such that non-existence of property is same as false anyway? i.e. `mule.http.disableResponseStreaming`

src/main/java/org/mule/extension/http/internal/HttpConnectorConstants.java
```diff
@@ -20,5 +20,6 @@
   String CONNECTOR_OVERRIDES = "Connector Overrides";
   String TLS_CONFIGURATION = "TLS Configuration";
   String AUTHENTICATION = "Authentication";
+  String STREAM_RESPONSE_PROPERTY = "mule.http.streamResponse";
```
dfeist: Is `org.mule.runtime.core.api.config.MuleProperties#SYSTEM_PROPERTY_PREFIX` visable to use?

PR #40, opened 2017-07-02
MULE-10401: Remove HTTP parsing mechanism.

afelisatti:


src/main/java/org/mule/extension/http/internal/request/HttpResponseToResult.java
```diff
@@ -89,24 +70,6 @@ public HttpResponseToResult(HttpRequesterCookieConfig config, Boolean parseRespo
     Charset encoding = responseMediaType.getCharset().get();
 
     Mono<?> payload = just(entity.getContent());
-    if (responseContentType != null && parseResponse) {
-      if (entity.isComposed()) {
-        responseMediaType = getJavaMediaType(encoding);
-        // Given we need to read whole payload in this scenario, do this using IO scheduler for avoid deadlock
```
elrodro83: great we could remove this!

PR #41, opened 2017-07-03
MULE-9682 - Define the list of objects that we will allow to inject i…

pablolagreca:
…n the domain objects


src/main/java/org/mule/extension/http/internal/listener/HttpListener.java
```diff
@@ -111,6 +111,9 @@
   private static final HttpListenerErrorResponseBuilder NULL_ERROR_RESPONSE = new HttpListenerErrorResponseBuilder();
 
   @Inject
+  private TransformationService transformationService;
+
+  @Inject
   private MuleContext muleContext;
```
marianogonzalez: org.mule.extension.http.internal.request.HttpRequestFactory#createRequestEntityFromPayload still uses is for transformation. If you replace that one with the TransformationService, pretty sure you can replace it here as well

pablolagreca: MuleContext is not used for that purpose in that method. It Is used for getting the default encoding.

PR #42, opened 2017-07-09
MULE-10878: HTTP does not support multiple headers/query params (same name)

afelisatti:


src/main/java/org/mule/extension/http/api/error/HttpErrorMessageGenerator.java
```diff
@@ -0,0 +1,61 @@
...
+      stringBuilder.append(" with status code ").append(statusCode);
+    }
+    stringBuilder.append(".");
+
+    return stringBuilder.toString();
+  }
+
+  private StringBuilder getBaseMessage(HttpRequest request) {
+    StringBuilder stringBuilder = new StringBuilder();
+    stringBuilder.append("HTTP ").append(request.getMethod()).append(" on resource ").append(request.getUri()).append(" failed");
```
elrodro83: isn't the word `resource` redundant in this context?

elrodro83: quote the url

src/main/java/org/mule/extension/http/api/error/HttpErrorMessageGenerator.java
```diff
@@ -0,0 +1,61 @@
...
+   * and the received status code for all errors.
+   *
+   * @param request the {@link HttpRequest} that resulted in the failure response
+   * @param statusCode the status code received as a response
+   * @return the message to be used on exceptions
+   */
+  public String createFrom(HttpRequest request, int statusCode) {
+    StringBuilder stringBuilder = getBaseMessage(request);
+    Optional<HttpError> httpError = getErrorByCode(statusCode);
+    if (httpError.isPresent()) {
```
elrodro83: use `map` and `orElse` from `Optional`

src/main/java/org/mule/extension/http/api/error/HttpErrorMessageGenerator.java
```diff
@@ -0,0 +1,61 @@
...
+      stringBuilder.append(": ").append(httpError.get().getErrorMessage(request)).append(" (").append(statusCode).append(")");
```
elrodro83: can it happen that the server returns a code with a different message than the rfc specifies? If so, that should be logged too.

src/main/java/org/mule/extension/http/api/error/HttpErrorMessageGenerator.java
afelisatti: Not really, it's just a way to make clear that the issue might be the path.

src/main/java/org/mule/extension/http/api/error/HttpErrorMessageGenerator.java
afelisatti: I'm not sure about that. Our messages should be consistent considering only the status code since that's what can be configured as a failure. If not one HTTP:NOT_FOUND could have a different message than another. Plus the messages are customised on purposed for some of the errors and in any case the actual response data is available in the error message field.

PR #44, opened 2017-07-11
MULE-13031: Move HTTP test from integrations to connector

afelisatti:


src/test/java/org/mule/test/http/functional/requester/HttpRequestProxyConfigTestCase.java
```diff
@@ -101,28 +89,9 @@ public void stopMockProxy() throws Exception {
 
   @Test
   public void testProxy() throws Exception {
-    checkProxyConfig();
```
elrodro83: why remove this check?

src/test/java/org/mule/test/http/functional/BasicHttpTestCase.java
```diff
@@ -20,114 +21,34 @@
...
-    response.setContentType("text/html");
-    response.setStatus(HttpServletResponse.SC_OK);
-    response.getWriter().print("WOW");
-  }
-
   @Test
   public void sendsRequest() throws Exception {
     Event response = flowRunner("client").withPayload("PEPE").run();
-    assertThat(response.getMessage().getPayload().getValue(), is("WOW"));
+    assertThat(response.getMessage().getPayload().getValue(), is(DEFAULT_RESPONSE));
```
elrodro83: is this change ok?

afelisatti: Yes, cause I'm now extending from the abstract HTTP request test case that has another handler to avoid all that duplicate jetty server code.

src/test/java/org/mule/test/http/functional/requester/HttpRequestProxyConfigTestCase.java
afelisatti: To keep this checks that were basically making sure the config had been created with the specified values I had to expose the HttpRequestProvider and all that package. Since the SDK tests must be validating that already and I don't want to compromise the isolation testing as a whole, I decided to remove this.

PR #45, opened 2017-07-12
MULE-12994 Support flag -Dmule.timeout.disable=true in HTTP Listeners

dfmacias:


src/test/java/org/mule/test/http/functional/listener/HttpListenerDisableTimeoutsConfigTestCase.java
```diff
@@ -0,0 +1,52 @@
...
+import static org.hamcrest.Matchers.is;
+import static org.mule.runtime.core.api.config.MuleProperties.SYSTEM_PROPERTY_PREFIX;
+
+public class HttpListenerDisableTimeoutsConfigTestCase extends AbstractHttpTestCase {
+
+  @Rule
+  public DynamicPort port = new DynamicPort("port");
+
+  @Rule
+  public SystemProperty disableTimeouts = new SystemProperty(SYSTEM_PROPERTY_PREFIX + "timeout.disable", "true");
```
pablokraan: There should be a constant somewhere for "timeout.disable" and it must be referenced here

PR #61, opened 2017-07-21
MULE-13162: Fix NullPointer thrown when resolving HTTP request policy…

balbifm:
… parameters


src/main/java/org/mule/extension/http/api/policy/HttpPolicyRequestParametersTransformer.java
```diff
@@ -36,11 +36,13 @@ public boolean supports(ComponentIdentifier componentIdentifier) {
   @Override
   public Message fromParametersToMessage(Map<String, Object> parameters) {
     String path = (String) parameters.get("path");
-    TypedValue<Object> body = (TypedValue<Object>) parameters.get("body");
+    TypedValue<Object> body = (TypedValue<Object>) parameters.getOrDefault("body", TypedValue.of(null));
```
afelisatti: Is this correct? The default is `#[payload]` for the response and `#[error.description]` for error response.

PR #85, opened 2017-08-15
MULE-13009: Inconsistent error propagation in extension operations depending on operation type

juandesi:


src/test/java/org/mule/test/http/functional/requester/HttpRequestErrorMappingsTestCase.java
```diff
@@ -64,14 +61,13 @@ public void matchingMappedRequest() throws Exception {
   @Test
   @Description("Verifies that an unmapped error is handled as ANY.")
   public void noMatchingMappedRequest() throws Exception {
-    verify("complexMapping", UNMATCHED_ERROR_MESSAGE, "Potato!");
+    verify("complexMapping", UNMATCHED_ERROR_MESSAGE, this);
```
marianogonzalez: `this` for headers?

juandesi: we need to pass anything that makes DW fail. "Potato!" does not fail anymore :)

afelisatti: Such a shame... "Potato!" was funnier.

PR #94, opened 2017-09-01
MULE-13429: Separate wrapping logic from ErrorMessageAwareException

afelisatti:


src/test/java/org/mule/test/http/functional/listener/HttpListenerErrorInterpretationTestCase.java
```diff
@@ -248,10 +248,20 @@ public ErrorException() {
...
-      super(new WrapperErrorMessageAwareException(Message.builder()
-          .value(ERROR)
-          .attributesValue(attributesToSend)
-          .build(), new IOException(OOPS)), errorToThrow);
+      super(new InnerErrorMessageException(OOPS), errorToThrow);
+    }
+
+  }
+
+  public static class InnerErrorMessageException extends IOException implements ErrorMessageAwareException {
```
pablokraan: Needs to be public?

PR #99, opened 2017-09-08
MULE-13363: Allow access to session properties from privileged API

elrodro83:
* Change security filter return type to 
* Remove some deprecated usages
* Event interface split


src/main/java/org/mule/extension/http/internal/listener/HttpListener.java
```diff
@@ -332,8 +332,6 @@ public void handleRequest(HttpRequestContext requestContext, HttpResponseReadyCa
         } catch (RuntimeException e) {
           LOGGER.warn("Exception occurred processing request:", e);
           sendErrorResponse(INTERNAL_SERVER_ERROR, SERVER_PROBLEM, responseCallback);
-        } finally {
-          setCurrentEvent(null);
```
pablokraan: Why is not needed anymore?

elrodro83: This seems to be a leftover from 3.x. The reactive pipeline is responsible of doing this cleanup now, not the connector.

PR #105, opened 2017-09-13
MULE-13401: Adapt multipart test once DW bugs are fixed

afelisatti:
This changes require a DW fix.


afelisatti:
Bug has been fix.

PR #108, opened 2017-09-18
MULE-13563: Error Handling with policies - Setting default statusCode and payload on success and failure

FedeAmdan:


src/main/java/org/mule/extension/http/api/policy/HttpListenerPolicyParametersTransformer.java
```diff
@@ -66,33 +66,37 @@ public Message fromFailureResponseParametersToMessage(Map<String, Object> parame
...
+                                                          String responseBuilderParameterName, Message message,
+                                                          int defaultStatusCode) {
     ImmutableMap.Builder<String, Object> mapBuilder =
         ImmutableMap.<String, Object>builder().put(responseBuilderParameterName, httpListenerResponseBuilder);
     if (message.getAttributes().getValue() instanceof HttpResponseAttributes) {
       HttpResponseAttributes httpResponseAttributes = (HttpResponseAttributes) message.getAttributes().getValue();
       httpListenerResponseBuilder.setBody(message.getPayload());
       httpListenerResponseBuilder.setHeaders(httpResponseAttributes.getHeaders());
-      httpListenerResponseBuilder.setStatusCode(httpResponseAttributes.getStatusCode());
+      httpListenerResponseBuilder.setStatusCode(httpResponseAttributes.getStatusCode() == 0 ? defaultStatusCode
```
afelisatti: So this depends on statusCode remaining undefined? Wouldn't it be best to distinguished success from error attributes and give those the defaults? Or even better, settling on 200 and redefining it to 500 in case of errors.

src/main/java/org/mule/extension/http/api/policy/HttpListenerPolicyParametersTransformer.java
```diff
@@ -38,8 +38,9 @@ public boolean supports(ComponentIdentifier componentIdentifier) {
 
   @Override
   public Message fromSuccessResponseParametersToMessage(Map<String, Object> parameters) {
-    HttpListenerResponseBuilder responseBuilder = (HttpListenerResponseBuilder) parameters.get("response");
-    return responseParametersToMessage(responseBuilder, 200);
+    HttpListenerResponseBuilder responseBuilder =
+        (HttpListenerResponseBuilder) parameters.get(ResponseType.SUCCESS.getResponseBuilderParameterName());
```
afelisatti: static imports all around for these enums

src/main/java/org/mule/extension/http/api/policy/HttpListenerPolicyParametersTransformer.java
```diff
@@ -60,42 +61,79 @@ private Message responseParametersToMessage(HttpListenerResponseBuilder response
...
       return mapBuilder.build();
     } else {
-      httpListenerResponseBuilder.setBody(message.getPayload());
+      if (ResponseType.SUCCESS.equals(responseType)) {
+        httpListenerResponseBuilder.setBody(message.getPayload());
+        httpListenerResponseBuilder
+            .setStatusCode(httpListenerResponseBuilder.getStatusCode() == null ? responseType.getStatusCode()
+                : httpListenerResponseBuilder.getStatusCode());
+      } else {
+        httpListenerResponseBuilder.setStatusCode(responseType.getStatusCode());
```
afelisatti: Why not consider the builder status code in this scenario as above?

PR #111, opened 2017-09-20
MULE-12716: Remove registries from API

elrodro83:


src/test/java/org/mule/test/http/functional/requester/HttpRequestLifecycleTestCase.java
```diff
@@ -28,48 +32,63 @@
   @Rule
   public ExpectedException expectedException = ExpectedException.none();
 
+  @Inject
+  private Registry registry;
+
+  //  @Inject
```
pablokraan: Remove commented code

PR #116, opened 2017-09-23
MULE-12716: Remove registries from API

elrodro83:


src/test/java/org/mule/test/http/functional/HttpRequesterTlsRestrictedProtocolsAndCiphersTestCase.java
```diff
@@ -21,4 +21,9 @@
   protected String getConfigFile() {
     return "http-request-restricted-protocols-ciphers-config.xml";
   }
+
+  @Override
+  protected boolean doTestClassInjection() {
+    return true;
```
pablokraan: Is not this the default value on the functional tests?

elrodro83: The default is set for `MuleArtifactFunctionalTestCase`, this one extends from another test class (there is a TODO to change it)

PR #139, opened 2017-10-20
MULE-11196: return a 503 response when listener is out of workers

alepulver:


src/test/java/org/mule/test/http/functional/HttpListenerOverloadTestCase.java
```diff
@@ -0,0 +1,147 @@
...
+import org.apache.http.HttpResponse;
+import org.apache.http.client.fluent.Executor;
+import org.apache.http.client.fluent.Request;
+import org.apache.http.impl.client.HttpClientBuilder;
+import org.apache.http.impl.conn.PoolingHttpClientConnectionManager;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+
+@Feature(HTTP_EXTENSION)
```
afelisatti: No need for this since it's in the super class, add a `Story` annotation though so we can track this is for the overload scenario.

src/test/java/org/mule/test/http/functional/HttpListenerOverloadTestCase.java
```diff
@@ -0,0 +1,147 @@
...
+      for (int i = 0; i < 5; i++) {
+        final HttpResponse response = Request.Get(url).connectTimeout(RECEIVE_TIMEOUT).execute().returnResponse();
+        assertThat(response.getStatusLine().getStatusCode(), is(SERVICE_UNAVAILABLE.getStatusCode()));
+        assertThat(response.getStatusLine().getReasonPhrase(), is(SERVICE_UNAVAILABLE.getReasonPhrase()));
+        assertThat(IOUtils.toString(response.getEntity().getContent()), is("Scheduler unavailable"));
+      }
+
+    } finally {
+      keepProcessorsActive.release();
+      if (!activeProcessorsFinished.await(15, TimeUnit.SECONDS)) {
```
afelisatti: That a very large timeout, have you validated whether it can be lowered? You could use the flakiness detector runner to check.

src/test/java/org/mule/test/http/functional/HttpListenerServiceUnavailableTestCase.java
```diff
@@ -0,0 +1,45 @@
...
+import org.mule.runtime.core.api.util.IOUtils;
+import org.mule.tck.junit4.rule.DynamicPort;
+
+import io.qameta.allure.Feature;
+import org.apache.http.HttpResponse;
+import org.apache.http.client.fluent.Request;
+import org.junit.Rule;
+import org.junit.Test;
+
+@Feature(HTTP_EXTENSION)
```
afelisatti: Same.

PR #143, opened 2017-10-31
MULE-13896: Improve how URIs are created

mbuchwald:


src/main/java/org/mule/extension/http/internal/listener/HttpRequestAttributesBuilder.java
```diff
@@ -44,9 +47,15 @@ public HttpRequestAttributes build() {
...
     String scheme = requestContext.getScheme();
     String method = request.getMethod();
-    String uri = request.getUri();
-    String path = extractPath(uri);
-    String queryString = extractQueryParams(uri);
+
+    URI uri = request.getUri();
+    String path = uri.getPath();
+    String uriString = path;
+    Optional<String> query = ofNullable(uri.getQuery());
```
elrodro83: extract to a method

elrodro83: What's the point of generating an Optional just to check `isPresent` afterwards?

src/test/java/org/mule/test/http/functional/listener/HttpListenerConfigFunctionalTestCase.java
```diff
@@ -99,7 +99,7 @@ public void listenerConfigOverridesListenerConfig() throws Exception {
   public void noListenerConfig() throws Exception {
     final String url = format("http://localhost:%s/noListener", noListenerConfigPort.getNumber());
     final HttpResponse httpResponse = callAndAssertStatus(url, NOT_FOUND);
-    assertThat(IOUtils.toString(httpResponse.getEntity().getContent()), is(format(NO_LISTENER_ENTITY_FORMAT, "/noListener")));
+    assertThat(IOUtils.toString(httpResponse.getEntity().getContent()), is(format(NO_LISTENER_ENTITY_FORMAT, url)));
```
elrodro83: Same as in https://github.com/mulesoft/mule-http-service/pull/56#discussion-diff-148074617R77

PR #150, opened 2017-11-22
MULE-13910: differentiate source and flow overload conditions

alepulver:


src/test/java/org/mule/test/http/functional/HttpListenerSourceOverloadTestCase.java
```diff
@@ -100,7 +100,9 @@ public void overloadScenario() throws Exception {
     }
 
     if (accumulatedErrors.size() > 0) {
-      fail(accumulatedErrors.stream().limit(10).collect(Collectors.toList()).toString());
+      String errorMessage =
+          String.join("\n", accumulatedErrors.stream().limit(10).map(x -> x.toString()).collect(Collectors.toList()));
```
elrodro83: static import

dfeist:
@alepulver Can you update the logic here based on this change?  `instanceof AbstractPipeline` should no longer be needed now.

PR #151, opened 2017-11-22
MULE-14035: Automatic retry on remotely closed exception.

facundovs:


src/test/resources/http-retry-policy-config.xml
```diff
@@ -0,0 +1,21 @@
...
+      xsi:schemaLocation="
+               http://www.mulesoft.org/schema/mule/core http://www.mulesoft.org/schema/mule/core/current/mule.xsd
+               http://www.mulesoft.org/schema/mule/http http://www.mulesoft.org/schema/mule/http/current/mule-http.xsd">
+
+    <http:request-config name="requestConfig" >
+        <http:request-connection host="localhost" port="${httpPort}" />
+    </http:request-config>
+
+    <flow name="retryIdempotentMethod">
+        <http:request config-ref="requestConfig" method="GET"  path="test" />
```
afelisatti: use one flow and parameterise the method through a variable instead

src/test/java/org/mule/test/http/functional/requester/AbstractHttpRetryRequestTestCase.java
```diff
@@ -0,0 +1,83 @@
...
+    retryAttemptsField.setAccessible(true);
+    retryAttemptsField.setInt(null, getNumberOfRetries());
+  }
+
+  void runIdempotentFlow() throws Exception {
+    runIdempotentFlow(DEFAULT_RETRY_ATTEMPTS);
+  }
+
+  void runIdempotentFlow(int numberOfRetryExpected) throws Exception {
+    TestServerSocket testServerSocket = new TestServerSocket(port.getNumber(), numberOfRetryExpected + 1);
```
afelisatti: When are you disposing this threads?

src/test/java/org/mule/test/http/functional/requester/HttpRetryRequestAttemptsTestCase.java
```diff
@@ -0,0 +1,47 @@
...
+    int customRetryRequests = 2;
+    return new Object[] {notRetryRequests, customRetryRequests};
+  }
+
+  public HttpRetryRequestAttemptsTestCase(Integer numberOfRetries) {
+    this.numberOfRetries = numberOfRetries;
+  }
+
+  @Test
+  public void customRetryRequestAttemptsIdempotentMethod() throws Exception {
```
afelisatti: Looks like the tests could be defined in the abstract class as well

src/test/java/org/mule/test/http/functional/requester/AbstractHttpRetryRequestTestCase.java
```diff
@@ -0,0 +1,83 @@
...
+  public ExpectedException expectedException = ExpectedException.none();
+
+  @Override
+  protected String getConfigFile() {
+    return "http-retry-policy-config.xml";
+  }
+
+  @Before
+  public void setUp() throws Exception {
+    expectedException.expectCause(REMOTELY_CLOSE_CAUSE_MATCHER);
```
afelisatti: I don't understand why you are setting the expected exception in a before. It should be used right before the method call that will fail.

src/test/java/org/mule/test/http/functional/requester/AbstractHttpRetryRequestTestCase.java
```diff
@@ -0,0 +1,83 @@
...
+  @Rule
+  public ExpectedException expectedException = ExpectedException.none();
+
+  @Override
+  protected String getConfigFile() {
+    return "http-retry-policy-config.xml";
+  }
+
+  @Before
+  public void setUp() throws Exception {
```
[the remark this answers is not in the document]
facundovs: I can't. The system property that sets the retry attempts is read statically.
As all the tests run in the same JVM, I need a way to redefine the value of the system property.

src/test/java/org/mule/test/http/functional/requester/AbstractHttpRetryRequestTestCase.java
facundovs: All the tests share the same exception cause. I thought it would be a good manner to avoid the same line in all the tests.

src/test/java/org/mule/test/http/functional/requester/AbstractHttpRetryRequestTestCase.java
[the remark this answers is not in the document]
afelisatti: Then don't read it statically.

src/test/java/org/mule/test/http/functional/requester/AbstractHttpRetryRequestTestCase.java
afelisatti: The thing is that if something else you don't expect to fail does it will be covered by that matcher instead of simply bubbling up.

src/test/java/org/mule/test/http/functional/requester/AbstractHttpRetryRequestTestCase.java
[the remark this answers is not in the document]
facundovs: Reading the system property only once was a code review suggestion by @dfeist in 3.x

src/test/java/org/mule/test/http/functional/requester/HttpRequestProxyConfigTestCase.java
```diff
@@ -98,7 +98,7 @@ public void testProxy() throws Exception {
   private void ensureRequestGoesThroughProxy(String flowName) throws Exception {
     // Request should go through the proxy.
     flowRunner(flowName).withPayload(TEST_MESSAGE).runExpectingException(allOf(instanceOf(HttpRequestFailedException.class),
-                                                                               hasMessage(containsString("Remotely closed"))));
+                                                                               hasMessage(containsString("Connection refused"))));
```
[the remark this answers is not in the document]
facundovs: Now , the request is retried, and the exception message changed. This same test was modified in 3.x. 
Other option is to turn off the retry policy to get the same exception message.

PR #172, opened 2017-12-21
MULE-14160: Add marker interface for exposing the proxy configuration…

facundovs:
… as an imported type.


src/main/java/org/mule/extension/http/api/request/proxy/HttpProxyConfig.java
```diff
@@ -0,0 +1,19 @@
...
+package org.mule.extension.http.api.request.proxy;
+
+import org.mule.runtime.http.api.client.proxy.ProxyConfig;
+
+/**
+ * Marker interface for exposing the proxy configuration as an imported type.
+ */
+public interface HttpProxyConfig extends ProxyConfig {
+
+  interface HttpNtlmProxyConfig extends NtlmProxyConfig {
```
elrodro83: shouldn't this interface also extend `HttpProxyConfig`?

PR #173, opened 2017-12-22
MULE-14271: HTTP Connector should use FAIL BackpressueStrategy by def…

marianogonzalez:
…ault

### DIFF
--- a/src/main/java/org/mule/extension/http/internal/listener/HttpListener.java
+++ b/src/main/java/org/mule/extension/http/internal/listener/HttpListener.java
@@ -24,7 +24,9 @@ import static org.mule.runtime.core.api.lifecycle.LifecycleUtils.startIfNeeded;
 import static org.mule.runtime.core.api.util.SystemUtils.getDefaultEncoding;
 import static org.mule.runtime.extension.api.annotation.param.MediaType.ANY;
 import static org.mule.runtime.extension.api.annotation.param.display.Placement.ADVANCED_TAB;
+import static org.mule.runtime.extension.api.runtime.source.BackPressureMode.DROP;
 import static org.mule.runtime.http.api.HttpConstants.HttpStatus.BAD_REQUEST;
+import static org.mule.runtime.http.api.HttpConstants.HttpStatus.DROPPED;
 import static org.mule.runtime.http.api.HttpConstants.HttpStatus.INTERNAL_SERVER_ERROR;
 import static org.mule.runtime.http.api.HttpConstants.HttpStatus.SERVICE_UNAVAILABLE;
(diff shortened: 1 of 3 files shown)


src/main/java/org/mule/extension/http/internal/listener/HttpListener.java
```diff
@@ -233,6 +250,31 @@ private void sendErrorResponse(HttpListenerErrorResponseBuilder errorResponse,
     responseCallback.responseReady(response, getResponseFailureCallback(responseCallback, completionCallback));
   }
 
+  private void sendBackPressureResponse(BackPressureContext ctx, SourceCompletionCallback completionCallback) {
```
dfeist: In theory, there was already logic somewhere sending  `SERVICE_UNAVAILABLE` but it doesn't seem you have reused it or deleted it either.

src/test/java/org/mule/test/http/functional/listener/HttpBackPressureTestCase.java
```diff
@@ -0,0 +1,101 @@
...
+import org.junit.Rule;
+import org.junit.Test;
+
+public class HttpBackPressureTestCase extends AbstractHttpTestCase {
+
+  @Rule
+  public DynamicPort listenPort = new DynamicPort("port");
+
+  @Inject
+  private HttpService httpService;
```
afelisatti: Why was this necessary? `AbstractHttpTestCase` has a client available.

src/main/java/org/mule/extension/http/internal/listener/HttpListener.java
afelisatti: The logic seems to be in `createFailureResponseBuilder`

marianogonzalez: this logic is similar to createFailureResponseBuilder() but with significant differences. Notice that I don't have a base error builder, the Error object will not have a Message associated with it, etc.

src/test/java/org/mule/test/http/functional/listener/HttpBackPressureTestCase.java
marianogonzalez: simply because I didn't know that

src/main/java/org/mule/extension/http/internal/listener/HttpListener.java
afelisatti: We need to figure what of that won't happen now and remove that code.

dfeist: I think it's just remove:
```
if (error.getErrorType().equals(sourceOverloadErrorType)) {
         failureResponseBuilder = createDefaultFailureResponseBuilder(error, SERVICE_UNAVAILABLE);
       } 
```

PR #186, opened 2018-01-15
MULE-14437: Exported packages are not communicated when loading an ex…

estebanwasinger:
…tension with the AST Loader


afelisatti:
:crossed_fingers:

PR #200, opened 2018-01-25
MULE-14530: Http: Allow requester to keep provided headers case

elrodro83:


src/test/java/org/mule/test/http/functional/HttpHeaderCaseTestCase.java
```diff
@@ -37,14 +52,86 @@ protected String getConfigFile() {
...
   }
+
+  @Test
+  public void proxyPreservesRequestHeaderCase() throws ClientProtocolException, IOException {
+    HttpResponse response = Request.Get(format("http://localhost:%s/proxyRequest", port.getNumber()))
+        .execute().returnResponse();
+
+    assertThat(response.getStatusLine().getStatusCode(), is(OK.getStatusCode()));
+    Set<String> headerNames = stream(response.getAllHeaders()).map(h -> h.getName()).collect(toSet());
+    assertThat(headerNames, hasItems("pRoXyHeAdEr"));
```
afelisatti: This should be sent here and only copied in the requester.

PR #239, opened 2018-07-06
MULE-15174:  Add global ResponseValidator in the requester configurat…

alegmarra:
…ion.


src/test/resources/http-request-status-codes-config.xml
```diff
@@ -19,18 +26,15 @@
     </flow>
 
     <flow name="success">
-        <http:request config-ref="requestConfig" path="status">
+        <http:request config-ref="requestConfigWithDefaultSuccessValidator" path="status">
```
afelisatti: This looks like it's reducing the coverage, shouldn't you add tests here instead of override the existing ones?

alegmarra:
this should be merged in 1.3.x

PR #249, opened 2018-08-09
MULE-15484: Add "proxyRequest" to HttpRequestAttributes to assess performance

lucianoRM:


src/main/java/org/mule/extension/http/api/HttpRequestAttributesBuilder.java
```diff
@@ -139,7 +142,36 @@ public HttpRequestAttributes build() {
...
+    this.proxyRequestPath = resolveProxyRequestPath();
+    return new HttpRequestAttributes(headers, listenerPath, relativePath, proxyRequestPath, version, scheme, method, requestPath,
+                                     requestUri,
                                      queryString, queryParams, uriParams, localAddress, remoteAddress, clientCertificate);
   }
+
+  private String resolveProxyRequestPath() {
+    byte[] listenerPathBytes = listenerPath.getBytes();
+    byte[] requestPathBytes = requestPath.getBytes();
+    int listenerPathIndex = 0;
```
afelisatti: I think these two vars would make more sense as `listenerCurrentSlashIndex` or something that relates the index to the fact that it's pointing to a slash position

src/main/java/org/mule/extension/http/api/HttpRequestAttributesBuilder.java
```diff
@@ -139,7 +142,36 @@ public HttpRequestAttributes build() {
...
+    } catch (ArrayIndexOutOfBoundsException e) {
+      //We could reach an error here if paths are not formed consistently. E.g: The listenerPath does not ends with *.
+      //In that case we return null and let other logic parse the path as if this never existed
+      return null;
+    }
+    return new String(copyOfRange(requestPathBytes, requestPathIndex - 1, requestPathBytes.length));
+  }
+
+  private int iterateUntilSlash(byte[] bytes, int position) {
+    while (bytes[position] != '/') {
```
afelisatti: extract to a constant

src/main/java/org/mule/extension/http/api/HttpRequestAttributesBuilder.java
```diff
@@ -139,7 +142,36 @@ public HttpRequestAttributes build() {
     requireNonNull(requestUri, "Request URI cannot be null.");
     requireNonNull(localAddress, "Local address cannot be null.");
     requireNonNull(remoteAddress, "Remote address cannot be null.");
-    return new HttpRequestAttributes(headers, listenerPath, relativePath, version, scheme, method, requestPath, requestUri,
+    this.proxyRequestPath = resolveProxyRequestPath();
```
afelisatti: I think you are recalculating the proxyRequestPath for each `build` but it may already be present. We should find a way to avoid this. Perhaps calculating it whenever the listenerPath or requestPath change and the other is present?

PR #261, opened 2018-09-20
MULE-15722: HTTP Requester is sending content-type even if no payload…

afelisatti:
… is sent

### DIFF
--- a/src/main/java/org/mule/extension/http/internal/request/HttpRequestFactory.java
+++ b/src/main/java/org/mule/extension/http/internal/request/HttpRequestFactory.java
@@ -23,7 +23,6 @@ import static org.mule.runtime.http.api.HttpHeaders.Names.COOKIE;
 import static org.mule.runtime.http.api.HttpHeaders.Names.TRANSFER_ENCODING;
 import static org.mule.runtime.http.api.HttpHeaders.Names.X_CORRELATION_ID;
 import static org.mule.runtime.http.api.HttpHeaders.Values.CHUNKED;
-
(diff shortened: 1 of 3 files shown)


src/main/java/org/mule/extension/http/internal/request/HttpRequestFactory.java
```diff
@@ -167,7 +159,16 @@ private HttpEntity createRequestEntity(HttpStreamingType streamingMode, HttpSend
     Optional<String> transferEncoding = requestBuilder.getHeaderValue(TRANSFER_ENCODING_HEADER);
     Optional<String> contentLength = requestBuilder.getHeaderValue(CONTENT_LENGTH_HEADER);
 
-    if (isEmptyBody(payload, resolvedMethod, sendBodyMode)) {
+    boolean emptyBody = isEmptyBody(payload, resolvedMethod, sendBodyMode);
+
+    MediaType mediaType = body.getDataType().getMediaType();
+    if (!requestBuilder.getHeaderValue(CONTENT_TYPE_HEADER).isPresent() && !emptyBody) {
```
lucianoRM: Why double "if" and not just another "&&"?

src/main/java/org/mule/extension/http/internal/request/HttpRequestFactory.java
```diff
@@ -167,7 +159,16 @@ private HttpEntity createRequestEntity(HttpStreamingType streamingMode, HttpSend
...
+      if (!MediaType.ANY.matches(mediaType)) {
```
lucianoRM: static import

PR #270, opened 2018-10-09
MULE-15786: lazyClientCertificate is not serializable in HttpAttributes

lucianoRM:


src/main/java/org/mule/extension/http/api/HttpRequestAttributes.java
```diff
@@ -171,7 +178,21 @@ public String getRemoteAddress() {
   }
 
   public Certificate getClientCertificate() {
-    return lazyClientCertificate.get();
+    if(this.lazyClientCertificate != null) {
```
marianogonzalez: format

marianogonzalez: also check for `clientCertificate == null`

src/main/java/org/mule/extension/http/api/HttpRequestAttributes.java
```diff
@@ -171,7 +178,21 @@ public String getRemoteAddress() {
...
+    return this.clientCertificate;
+  }
+
+  private void writeObject(ObjectOutputStream out) throws IOException {
+    //This will make the resolved client certificate be stored in the clientCertificate variable
+    getClientCertificate();
+    out.defaultWriteObject();
+  }
+
+  private void readObject(ObjectInputStream in) throws ClassNotFoundException, IOException {
```
afelisatti: I don't think you need this. You are counting on the write setting the correct `clientCertificate` value, so the supplier is irrelevant since that has been resolved.

src/test/java/org/mule/test/http/api/HttpRequestAttributesSerializationTestCase.java
```diff
@@ -0,0 +1,96 @@
...
+import org.mule.extension.http.api.HttpRequestAttributes;
+import org.mule.extension.http.api.HttpRequestAttributesBuilder;
+
+import java.security.cert.Certificate;
+
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+/**
+ * Tests for checking correct custom serialization in HttpRequestAttributes
```
afelisatti: Use allure for comments on tests.

src/main/java/org/mule/extension/http/api/HttpRequestAttributes.java
```diff
@@ -92,12 +93,23 @@
    */
   @Parameter
   @Optional
-  private final Certificate clientCertificate = null;
+  private transient Certificate clientCertificate;
```
marianogonzalez: how is this value ever obtained? the getter now returns the output of the lazyProvider

src/main/java/org/mule/extension/http/internal/certificate/CertificateProviderFactory.java
```diff
@@ -0,0 +1,44 @@
...
+      try {
+        return new DefaultCertificateProvider(certificateSupplier);
+      } catch (Exception e) {
+        LOGGER.warn("Errors while creating " + SERIALIZABLE_LAZY_VALUE_CLASS_NAME
+            + " by reflection, even when class in on classpath. Defaulting to compatibility implementation");
+      }
+    }
+    return new CompatibilityCertificateProvider(certificateSupplier);
+  }
+
```
marianogonzalez: remove empty lines

src/main/java/org/mule/extension/http/internal/certificate/CertificateProviderFactory.java
```diff
@@ -0,0 +1,44 @@
...
+  static {
+    isSerializableLazyValuePresent = isClassOnPath(SERIALIZABLE_LAZY_VALUE_CLASS_NAME, CertificateProvider.class);
+  }
+
+  public static CertificateProvider create(Supplier<Certificate> certificateSupplier) {
+    if (isSerializableLazyValuePresent) {
+      try {
+        return new DefaultCertificateProvider(certificateSupplier);
+      } catch (Exception e) {
+        LOGGER.warn("Errors while creating " + SERIALIZABLE_LAZY_VALUE_CLASS_NAME
```
marianogonzalez: not just warn. Blow up. Otherwise the system will continue to operate even though the certificates are lost

src/main/java/org/mule/extension/http/api/HttpRequestAttributes.java
```diff
@@ -93,7 +93,7 @@
    */
   @Parameter
   @Optional
-  private transient Certificate clientCertificate;
+  private final Certificate clientCertificate = null;
```
marianogonzalez: you can't do this. The SDK will try to set this value and DW might as well if this is used in a transformation. What you have to do is to store the result of the lazy supplier in this variable, making sure that when the supplier is resolved, the value is always available here. 

This also needs to have a getter which actually uses the field for DW compatibility

lucianoRM: This is how this was before all these changes. 
Besides, all parameters are declared final, why would they be changed by de SDK and DW? and how?
Why won't this be compatible with DW if it already has a getter that returns the value from the "lazy certificate"? I know that DW uses reflection when no getter is found for the attribute is field is trying to obtain, but if there is a getter present , it makes use of it.

PR #292, opened 2018-12-20
MULE-16246: Introduce rawRequestUri attribute

afelisatti:


src/test/resources/http-listener-message-properties-config.xml
```diff
@@ -14,6 +14,9 @@
     <http:listener-config name="listenerBaseConfig" basePath="api">
         <http:listener-connection host="localhost" port="${port2}"/>
     </http:listener-config>
+    <http:listener-config name="listenerEncodedBaseConfig" basePath="a p i">
```
elrodro83: this will be matched only for request that have this part properly encoded, right?

afelisatti: Yes, the test is hitting “a%20p%20i”

afelisatti: I added this test after realizing that listener paths are stored as defined (decoded if you will) so if we try to extract the base path from an encoded request path it wouldn’t work.

PR #298, opened 2019-01-19
MULE-16295: Add logger for resolved dynamic HTTP requests

afelisatti:


afelisatti:
@adobni please review and check whether this works for you. I wanted to add something more general. This way, when the connector logging level is set to DEBUG it will print out which method and full uri where the request will be sent to.

PR #308, opened 2019-03-12
MULE-16492: Reduce MultiMap copies.

afelisatti:


src/main/java/org/mule/extension/http/internal/listener/HttpResponseHeaderBuilder.java
```diff
@@ -36,27 +33,30 @@
...
-  private MultiMap<String, String> headers = new CaseInsensitiveMultiMap(!PRESERVE_HEADER_CASE);
+  private final HttpResponseBuilder responseBuilder;
+
+  public HttpResponseHeaderBuilder(HttpResponseBuilder responseBuilder) {
+    this.responseBuilder = responseBuilder;
+  }
 
   public void addHeader(String headerName, Collection<String> headerValue) {
-    if (headerValue.size() > 1 || headers.containsKey(headerName)) {
+    if (headerValue.size() > 1 || responseBuilder.getHeaderValue(headerName).isPresent()) {
```
marianogonzalez: if you're in the business of gaining milliseconds, isn't it best to add a `containsHeader(String):boolean` method on the response builder? In this way you save the `Optional` overhead

afelisatti: Yes, but that would only be available in the API version 1.2.0+.

PR #312, opened 2019-03-29
MULE-16632: Add missing MediaType to HTTP operations

aiannucci:


src/main/java/org/mule/extension/http/api/listener/intercepting/cors/CorsInterceptorWrapper.java
```diff
@@ -27,4 +29,19 @@
   public HttpListenerInterceptor getInterceptor() {
     return corsInterceptor;
   }
+
+  @Override
+  public boolean equals(Object o) {
+    if (this == o)
```
afelisatti: The formatter allowing this is a bug. Use braces here and below.

src/main/java/org/mule/extension/http/api/listener/intercepting/cors/CorsInterceptorWrapper.java
```diff
@@ -27,4 +29,19 @@
...
+      return true;
+    if (o == null || getClass() != o.getClass())
+      return false;
+    CorsInterceptorWrapper that = (CorsInterceptorWrapper) o;
```
afelisatti: We need to work with the API Gateway team which maintains this implementation so that they provide a proper hash and equals here, otherwise we will have issues all the same.

aiannucci: @balbifm what do you think?

aiannucci: @mtqp what do you think?

mtqp: We have released Cors Kernel 1.1.2, the HTTP Connector should point to that version now.

PR #314, opened 2019-04-04
MULE-16035: HTTP Connector does not implement equals/hash on POJOs

gsfernandes:
Missing equals/hashCode was still generating OOM Metaspace as CGLIB classes were created for the CachedConnectionProvider when using expressions on http:default-header


gsfernandes:
https://github.com/mulesoft/mule-http-connector/pull/315

PR #315, opened 2019-04-04
MULE-16652: HTTP Connector does not implement equals/hash on RequestH…

gsfernandes:
…eader POJO


gsfernandes:
Missing equals/hashCode was still generating OOM Metaspace as CGLIB classes were created for the CachedConnectionProvider when using expressions on http:default-header

PR #319, opened 2019-04-08
MULE-16669: Mule throws error Illegal group reference Exception when …

JuLuzzi:
…trying to pass $ into URI Parameter in HTTP request

### DIFF
--- a/src/main/java/org/mule/extension/http/api/request/builder/HttpRequesterRequestBuilder.java
+++ b/src/main/java/org/mule/extension/http/api/request/builder/HttpRequesterRequestBuilder.java
@@ -116,7 +116,7 @@ public class HttpRequesterRequestBuilder extends HttpMessageBuilder {
         throw new NullPointerException(format("Expression {%s} evaluated to null.", uriParamName));
       }
 
-      path = path.replaceAll("\\{" + uriParamName + "}", quoteReplacement(uriParamValue));
+      path = path.replaceAll("\\{" + uriParamName + "\\}", quoteReplacement(uriParamValue));
     }
     return path;
   }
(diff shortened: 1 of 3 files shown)


src/main/java/org/mule/extension/http/api/request/builder/HttpRequesterRequestBuilder.java
```diff
@@ -115,7 +116,7 @@ public String replaceUriParams(String path) {
         throw new NullPointerException(format("Expression {%s} evaluated to null.", uriParamName));
       }
 
-      path = path.replaceAll("\\{" + uriParamName + "\\}", uriParamValue);
+      path = path.replaceAll(format("\\{%s}", uriParamName), quoteReplacement(uriParamValue));
```
afelisatti: Why is the format necessary?

src/test/java/org/mule/test/http/functional/requester/HttpRequestUriParamsTestCase.java
```diff
@@ -63,4 +63,19 @@ public void sendsUriParamsIfNull() throws Exception {
     flowRunner("uriParamNull").run();
   }
 
+  @Test
+  public void uriParamsContainsReservedUriCharacter() throws Exception {
+    flowRunner("reservedUriCharacter").withPayload(AbstractMuleContextTestCase.TEST_MESSAGE)
```
afelisatti: static import

src/main/java/org/mule/extension/http/api/request/builder/HttpRequesterRequestBuilder.java
```diff
@@ -115,7 +116,7 @@ public String replaceUriParams(String path) {
         throw new NullPointerException(format("Expression {%s} evaluated to null.", uriParamName));
       }
 
-      path = path.replaceAll("\\{" + uriParamName + "\\}", uriParamValue);
+      path = path.replaceAll("\\{" + uriParamName + "}", quoteReplacement(uriParamValue));
```
afelisatti: I think you deleted extra `\\`

JuLuzzi: Only the opening brace needs to be scaped

PR #321, opened 2019-04-25
MULE-16756: Integrate all changes for HTTP request streaming

eze210:


pom.xml
```diff
@@ -5,9 +5,8 @@
     <parent>
         <groupId>org.mule.extensions</groupId>
         <artifactId>mule-core-modules-parent</artifactId>
-        <!-- TODO MULE-15249 if this is upgraded, change the Map to Set in HttpResponseHeaderBuilder -->
         <!-- TODO MULE-15803 if this is upgraded, change HttpRequestAttributes#lazyClientCertificate-->
-        <version>1.1.3</version>
+        <version>1.3.0-SNAPSHOT</version>
```
elrodro83: `1.2.0`: https://github.com/mulesoft/mule/pull/7837#discussion-diff-278545353R9

pom.xml
```diff
@@ -23,8 +22,8 @@
     <properties>
         <commonsCodecVersion>1.11</commonsCodecVersion>
         <fluentHcVersion>4.5.2</fluentHcVersion>
-        <grizzlyVersion>2.3.36-MULE-006</grizzlyVersion>
-        <grizzlyAhcVersion>1.14-MULE-007</grizzlyAhcVersion>
+        <grizzlyVersion>2.3.36-MULE-007-SNAPSHOT</grizzlyVersion>
```
elrodro83: do we really need to go through the snapshot here, or should we just release the dependency and change this dependency to a release directly?

pom.xml
```diff
@@ -23,8 +22,8 @@
...
+        <grizzlyAhcVersion>1.14-MULE-009-SNAPSHOT</grizzlyAhcVersion>
```
elrodro83: same as above

src/main/java/org/mule/extension/http/internal/listener/HttpResponseHeaderBuilder.java
```diff
@@ -16,21 +16,22 @@
...
-import java.util.Map;
+import java.util.Set;
+
 
 public class HttpResponseHeaderBuilder {
 
   // This is a Map instead of a Set because of MULE-15249
   // While already fixed, it might be present in older versions of the runtime used with this connector
-  private static final Map<String, String> uniqueHeadersNames;
+  private static final Set<String> uniqueHeadersNames;
```
elrodro83: remove the comment if no longer applies for 4.2.0

afelisatti:
This is not required. The changes are only to test things in 4.3.0-SNAPSHOT but the connector should remain untouched.

PR #330, opened 2019-06-03
MULE-16947: Add builders for HTTP attributes that currently don't hav…

balbifm:
…e one


src/main/java/org/mule/extension/http/api/HttpRequestAttributesBuilder.java
```diff
@@ -182,7 +182,7 @@ public HttpRequestAttributes build() {
     requireNonNull(requestUri, "Request URI cannot be null.");
     requireNonNull(localAddress, "Local address cannot be null.");
     requireNonNull(remoteAddress, "Remote address cannot be null.");
-    if (resolveMaskedRequestPath && listenerPath != null && requestPath != null) {
+    if (resolveMaskedRequestPath) {
```
afelisatti: wasn't this required to resolve the masked path?

balbifm: Oh, that was just a suggestion of IntelliJ :P. You have the requireNonNull of those two properties just above, so it is redundant to ask again.

PR #333, opened 2019-07-15
MULE-17110: NTLM authentication with dynamic credentials is successful despite credentials changes

plbalbi:


src/main/java/org/mule/extension/http/api/request/authentication/UsernamePasswordAuthentication.java
```diff
@@ -57,19 +58,23 @@ public boolean shouldRetry(Result<Object, HttpResponseAttributes> firstAttemptRe
...
+    return password.resolve();
   }
 
   @Override
   public boolean isPreemptive() {
     return preemptive;
   }
 
+  public boolean forceConnectionClose() {
+    return username.getExpression().isPresent() || password.getExpression().isPresent();
```
afelisatti: Please make double sure this works as expected when the value is static.

plbalbi: Just checked. Works.

PR #353, opened 2019-10-03
MULE-17525: Allow dynamic configuration for the proxy on the HTTP requester

eze210:


src/test/java/org/mule/test/http/functional/requester/proxy/HttpRequestProxyDynamicConfigTestCase.java
```diff
@@ -0,0 +1,61 @@
...
+import org.mule.tck.http.TestProxyServer;
+import org.mule.tck.junit4.rule.DynamicPort;
+import org.mule.test.http.functional.requester.AbstractHttpRequestTestCase;
+
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+
+public class HttpRequestProxyDynamicConfigTestCase extends AbstractHttpRequestTestCase {
```
afelisatti: I'd add another test using dynamic proxy authentication which was one of the original reasons behind the feature request.

PR #359, opened 2019-10-18
MULE-17670: HTTP ResponseBuilder should not return null headers

mbuchwald:


mbuchwald:
The `getHeader` had to be changed because SDK is injecting the NULL value. Also, `@NullSafe` annotation is not available for maps (https://github.com/mulesoft/mule/blob/mule-4.x/modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/validation/NullSafeModelValidator.java#L130)

src/main/java/org/mule/extension/http/api/listener/builder/HttpListenerErrorResponseBuilder.java
```diff
@@ -89,12 +89,12 @@ public void setBody(TypedValue<Object> body) {
...
   @Override
   public MultiMap<String, String> getHeaders() {
-    return headers;
+    return headers != null ? headers : emptyMultiMap();
   }
 
   @Override
   public void setHeaders(MultiMap<String, String> headers) {
-    this.headers = headers;
+    this.headers = headers != null ? headers : emptyMultiMap();
```
eze210: Could we add a warning in order to detect the user of this class which is setting a null as headers map?

mbuchwald: I don't think it's necessary. Moreover, I added the nullsafe because the problem was coming from the SDK injecting the null value, so the logging here wouldn't have helped us at all 🤷‍♂

PR #376, opened 2019-12-13
MULE-17508: Add test for error description content

eze210:


src/test/java/org/mule/test/http/functional/requester/HttpRequestExceptionMessagesTestCase.java
```diff
@@ -0,0 +1,38 @@
...
+
+  @Override
+  protected String getConfigFile() {
+    return "http-request-exception-messages.xml";
+  }
+
+  @Test
+  public void errorMessageIsLoadedFromCauseIfNull() throws Exception {
+    Exception exception = flowRunner("requesterFlow").runExpectingException();
+    assertThat(exception.getMessage(), not(isEmptyOrNullString()));
```
afelisatti: Why are you not asserting anything about the actual message content?

PR #377, opened 2019-12-19
MULE-17848: Add builder methods for http attribute classes

elrodro83:
* Also, add support for MTF tests

PR #378, opened 2019-12-19
MULE-17848: Add builder methods for http attribute classes (#377)

elrodro83:
* Also, add support for MTF tests

Cherry pick for 1.5.x, had to cherrypick some changes in `HttpResponseAttributesBuilder ` also.


src/main/java/org/mule/extension/http/internal/request/builder/HttpResponseAttributesBuilder.java
```diff
@@ -16,15 +20,36 @@
 
   HttpResponse response;
```
eze210: Before the commit, this attribute was used to get the others in build() method. Now it seems no longer needed

PR #394, opened 2020-01-15
MULE-17693: onErrorSendingResponse public method in anonymous class is not accessible via reflection

fsgonz:


fsgonz:
The way to replicate is using 4.1.5 and replacing the service for 1.2.4 and force a failed.

PR #409, opened 2020-02-06

eze210:


src/main/java/org/mule/extension/http/internal/listener/HttpListenerProvider.java
```diff
@@ -181,6 +186,7 @@ public void initialise() throws InitialisationException {
         .setPort(connectionParams.getPort())
         .setTlsContextFactory(tlsContext).setUsePersistentConnections(connectionParams.getUsePersistentConnections())
         .setConnectionIdleTimeout(connectionParams.getConnectionIdleTimeout())
+        .setSchedulerSupplier(() -> schedulerService.ioScheduler(SchedulerConfig.config().withName("http-listener-scheduler-io")))
```
elrodro83: add data to the scheduler of the name that can help disambiguate the schedulers when there are more than one listeners (protocol, host and port, for instance)

src/test/java/org/mule/test/PipedStreamGenerator.java
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  private static void writeChunkInStream(String chunk, PipedOutputStream stream) throws IOException {
+    stream.write(chunk.getBytes(), 0, chunk.length());
+    stream.flush();
+  }
+
+  private static String generateRandomString(int len) {
+    byte[] array = new byte[len];
+    new Random().nextBytes(array);
```
elrodro83: extract `new Random()` to a field

PR #410, opened 2020-02-06

eze210:


src/test/java/org/mule/test/PipedStreamGenerator.java
```diff
@@ -0,0 +1,73 @@
...
+
+  private static void writeChunkInStream(String chunk, PipedOutputStream stream) throws IOException {
+    stream.write(chunk.getBytes(), 0, chunk.length());
+    stream.flush();
+  }
+
+  private static String generateRandomString(int len) {
+    byte[] array = new byte[len];
+    random.nextBytes(array);
+    return new String(array, Charset.forName("UTF-8"));
```
elrodro83: Use StandardCharsets.UTF-8 instead

src/test/java/org/mule/test/PipedStreamGenerator.java
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  private static void writeChunkInStream(String chunk, PipedOutputStream stream) throws IOException {
+    stream.write(chunk.getBytes(), 0, chunk.length());
+    stream.flush();
+  }
+
+  private static String generateRandomString(int len) {
+    byte[] array = new byte[len];
+    random.nextBytes(array);
```
elrodro83: doesn't this have a method to generate a String?

src/test/java/org/mule/test/PipedStreamGenerator.java
```diff
@@ -0,0 +1,73 @@
...
+    return new String(array, Charset.forName("UTF-8"));
+  }
+
+  public static String writeChunkInStreams() throws IOException {
+    String chunk = generateRandomString(8192);
```
elrodro83: there is a KB constant somewhere to use here. streaming uses it if i recall correctly

src/test/java/org/mule/test/PipedStreamGenerator.java
```diff
@@ -0,0 +1,73 @@
...
+    for (PipedOutputStream stream : sources) {
+      writeChunkInStream(chunk, stream);
+    }
+    return "A chunk was written to " + sources.size() + " streams";
```
elrodro83: better have this method return an `int` with this info?

src/test/java/org/mule/test/PipedStreamGenerator.java
```diff
@@ -0,0 +1,73 @@
...
+
+  private static String generateRandomString(int len) {
+    byte[] array = new byte[len];
+    random.nextBytes(array);
+    return new String(array, Charset.forName("UTF-8"));
+  }
+
+  public static String writeChunkInStreams() throws IOException {
+    String chunk = generateRandomString(8192);
+    for (PipedOutputStream stream : sources) {
```
elrodro83: synchronize this block on `sources`

src/test/java/org/mule/test/PipedStreamGenerator.java
```diff
@@ -0,0 +1,73 @@
...
+      writeChunkInStream(chunk, stream);
+    }
+    return "A chunk was written to " + sources.size() + " streams";
+  }
+
+  public static void closeStreams() throws IOException {
+    for (PipedOutputStream stream : sources) {
```
elrodro83: synchronize this method on `sources`

src/test/munit/response-streaming/selectors-not-being-blocked-when-response-streaming.xml
```diff
@@ -0,0 +1,81 @@
...
+        </munit:validation>
+    </munit:test>
+    <sub-flow name="SpawnRequests" >
+        <choice doc:name="Choice" >
+            <when expression="#[vars.requestsNumber &gt; 0]">
+                <async doc:name="Request async" >
+                    <http:request method="GET" doc:name="Request" path="/stream" config-ref="HTTP_Request_config" requestStreamingMode="ALWAYS" />
+                </async>
+                <set-variable value="#[vars.requestsNumber-1]" doc:name="Decrease number of requests" variableName="requestsNumber" />
+                <flow-ref doc:name="Spawn N-1 requests" name="SpawnRequests" />
```
elrodro83: why not a `for-each` instead of using recursion?

src/test/java/org/mule/test/PipedStreamGenerator.java
eze210: Done

src/test/java/org/mule/test/PipedStreamGenerator.java
eze210: No. There is an apache method, but I would have to add a dependency just for this

src/test/java/org/mule/test/PipedStreamGenerator.java
eze210: Removed the return. I was using this info during manual testing, but it is no longer needed

src/test/java/org/mule/test/PipedStreamGenerator.java
eze210: Done

src/test/java/org/mule/test/PipedStreamGenerator.java
eze210: Done

src/test/java/org/mule/test/PipedStreamGenerator.java
eze210: Yes, it's the default size of the request streaming **buffer**. I don't know if it corresponds to use it here... What do you think?

src/test/munit/response-streaming/selectors-not-being-blocked-when-response-streaming.xml
eze210: If it walks like a duck and it quacks like a duck, then it must be a duck

eze210:
--validate

src/test/java/org/mule/test/PipedStreamGenerator.java
elrodro83: at least, extract a `KB` constant in this class, and multiply it by 8 here

eze210:
--validate

eze210:
--validate

PR #420, opened 2020-02-25
MULE-18108: Expose maskedPath as pointcut parameter in HTTP 1.x

balbifm:


src/main/java/org/mule/extension/http/api/policy/HttpListenerPolicyPointcutParameters.java
```diff
@@ -6,17 +6,37 @@
...
 
+  /**
+   * As long as the connector is compatible with runtime versions < 4.3.0, headers and masked request path need to be set using
+   * reflection since they were added in the API in 4.3.0
+   */
+  private static Method setHeadersMethod;
+  private static Method setMaskedRequestPathMethod;
+
+  static {
+    setHeadersMethod = ClassUtils.getMethod(HttpPolicyPointcutParameters.class, "setHeaders", null);
```
elrodro83: what if this is running with a runtime version 4.2.x or 4.1.x?
Isn't this missing the parameter type?

PR #422, opened 2020-03-03
Revert "MULE-18028: Use IO threads to handle HTTP response streaming …

fsgonz:


fsgonz:
I thought these branches were protected.

PR #423, opened 2020-03-06
MULE-18028: Use IO threads to handle HTTP response streaming (#410)

eze210:
(cherry picked from commit eb590ca71bad70859b93815a3f1bde0071effe28)


eze210:
--validate

PR #430, opened 2020-03-16
MULE-18109: Remove guava dependency from http-connector

elrodro83:


src/main/java/org/mule/extension/http/api/error/HttpError.java
```diff
@@ -132,11 +132,25 @@
...
+    httpRequestOperationErrors.add(UNAUTHORIZED);
+    httpRequestOperationErrors.add(METHOD_NOT_ALLOWED);
+    httpRequestOperationErrors.add(TOO_MANY_REQUESTS);
+    httpRequestOperationErrors.add(NOT_FOUND);
+    httpRequestOperationErrors.add(UNSUPPORTED_MEDIA_TYPE);
+    httpRequestOperationErrors.add(NOT_ACCEPTABLE);
+    httpRequestOperationErrors.add(INTERNAL_SERVER_ERROR);
+    httpRequestOperationErrors.add(SERVICE_UNAVAILABLE);
+
+    return unmodifiableSet(httpRequestOperationErrors);
```
eze210: What about make httpRequestOperationErrors a static method variable and creating it when it's null? This way you can avoid repeated additions, hash calculations... I don't know how many times is this method used

PR #431, opened 2020-03-17
MULE-17423: Remove runtime provided libraries from HTTP connector

elrodro83:


pom.xml
```diff
@@ -7,7 +7,7 @@
         <artifactId>mule-core-modules-parent</artifactId>
         <!-- TODO MULE-15249 if this is upgraded, change the Map to Set in HttpResponseHeaderBuilder -->
         <!-- TODO MULE-15803 if this is upgraded, change HttpRequestAttributes#lazyClientCertificate-->
-        <version>1.1.6</version>
+        <version>1.1.7</version>
```
eze210: Aren't the above MULEs needed with this upgrade?

elrodro83: no, those are for minor version upgrades

eze210: Ok

PR #432, opened 2020-03-18
MULE-18191: mule4 failed to stream rest of payload to client

szaffarano:


pom.xml
```diff
@@ -7,6 +7,7 @@
         <artifactId>mule-core-modules-parent</artifactId>
         <!-- TODO MULE-15249 if this is upgraded, change the Map to Set in HttpResponseHeaderBuilder -->
         <!-- TODO MULE-15803 if this is upgraded, change HttpRequestAttributes#lazyClientCertificate-->
+        <!-- TODO SE-14012 if this is upgraded, change HttpListener#onError -->
```
elrodro83: reference only MULE issues here

src/main/java/org/mule/extension/http/internal/listener/HttpListener.java
```diff
@@ -132,6 +132,10 @@
   private static final String REPEATED_HEADERS_LOG_FORMAT =
       "'X-Correlation-ID: {}' and 'MULE_CORRELATION_ID: {}' headers found. 'X-Correlation-ID' will be used.";
 
+  // Because of SE-14012, the service has to inform any error that implies lost the client connection
```
elrodro83: no SE references in this project

src/main/java/org/mule/extension/http/internal/listener/HttpListener.java
```diff
@@ -201,7 +205,9 @@ public void onError(@ParameterGroup(name = "Error Response",
                       Error error,
                       SourceCompletionCallback completionCallback) {
     try {
-      sendErrorResponse(errorResponse, callbackContext, error, completionCallback);
+      if (!error.getCause().getClass().getName().equals(CONNECTION_ALREADY_CLOSED_EXCEPTION)) {
```
elrodro83: from what i talked yesterday with @mlischetti , we may need a more generic approach in the runtime to handle this cases, otherwise this same situation can happen in other connectors trying to send data to closed connections

szaffarano:
I close this PR because we changed the solution y this fix is not necessary anymore.

PR #435, opened 2020-03-30

eze210:


src/test/java/org/mule/test/PipedStreamGenerator.java
```diff
@@ -44,7 +45,9 @@ private static String generateRandomString(int len) {
     return new String(array, UTF_8);
   }
 
-  public static void writeChunkInStreams() throws IOException {
+  public static void writeChunkInStreams() throws IOException, InterruptedException {
+    // The default tolerance before delegating response streaming in other thread is 50 ms.
+    sleep(50);
```
elrodro83: sleep for a little longer than the tolerance, to avoid flakiness

PR #436, opened 2020-04-01
MULE-18247: HTTP Requester should return proper 5xx errors

szaffarano:


src/main/java/org/mule/extension/http/api/error/HttpError.java
```diff
@@ -32,6 +35,12 @@
 
   TIMEOUT,
 
+  // represents any 4xx error
+  CLIENT_ERROR((Predicate<Integer>) statusCode -> statusCode >= 400 && statusCode < 500),
```
afelisatti, line 39: If an error handler matches `CLIENT_ERROR` and then `CLIENT_SECURITY` what would be the match for a 401 response?

src/test/resources/http-request-errors-config.xml
```diff
@@ -51,6 +51,18 @@
             <on-error-continue type="http:CONNECTIVITY">
                 <set-payload value="#[error.cause.message ++ ' connectivity']"/>
             </on-error-continue>
+            <on-error-continue type="http:GATEWAY_TIMEOUT">
+                <set-payload value="#[error.errorMessage.attributes.statusCode ++ ' Gateway Timeout']"/>
+            </on-error-continue>
+            <on-error-continue type="http:BAD_GATEWAY">
+                <set-payload value="#[error.errorMessage.attributes.statusCode ++ ' Bad Gateway']"/>
+            </on-error-continue>
+            <on-error-continue type="http:CLIENT_ERROR">
```
afelisatti, line 60: Perhaps `CLIENT_SIDE` and `SERVER_SIDE` would help avoid the `error` redundancy.

src/main/java/org/mule/extension/http/api/error/HttpError.java
szaffarano: CLIENT_ERROR only will match if there aren't any specific error defined, eg
```java
      try {
        error = HttpError.valueOf(status.name());
      } catch (Throwable e) {
        error = stream(HttpError.values()).filter(httpError -> httpError.isParentOf(statusCode))
            .findFirst()
            .orElse(null);
      }
```
Because there is an UNAUTHORIZED error already defined, this would be the match.

src/test/resources/http-request-errors-config.xml
szaffarano: +1

src/main/java/org/mule/extension/http/api/error/HttpError.java
```diff
@@ -132,7 +159,9 @@
       try {
         error = HttpError.valueOf(status.name());
       } catch (Throwable e) {
-        // Do nothing
+        error = stream(HttpError.values()).filter(httpError -> httpError.isParentOf(statusCode))
```
elrodro83, line 162: avoid relying on exceptions to run the alternate path.

src/main/java/org/mule/extension/http/api/error/HttpError.java
```diff
@@ -155,6 +184,10 @@
     return ofNullable(result);
   }
 
+  public boolean isParentOf(int statusCode) {
```
elrodro83, line 187: ad djavadoc

src/main/java/org/mule/extension/http/api/error/HttpError.java
elrodro83: this is still throwing an exception...

src/main/java/org/mule/extension/http/api/error/HttpError.java
```diff
@@ -132,7 +159,11 @@
       try {
         error = HttpError.valueOf(status.name());
```
elrodro83, line 160: this is still throwing an exception for the new scenario...

src/main/java/org/mule/extension/http/api/error/HttpError.java
```diff
@@ -155,6 +186,20 @@
...
+   * Returns whether the error is the parent of a given error represented by a {@code statusCode}. Parent means that the error
+   * belongs to the same hierarchy or kind of error. For example, 4xx are client errors, 5xx are server errors, etc, so a generic
+   * client error should be the parent of a more specific forbidden one (error code 403).
+   * 
+   * @param statusCode
+   * @return
+   *
+   * @since 1.6.0
+   */
+  public boolean isParentOf(int statusCode) {
```
elrodro83, line 199: maybe `category` is a more appropriate name for the relationship between these status code. This method should be named `isWithinCategory`.

src/main/java/org/mule/extension/http/api/error/HttpError.java
```diff
@@ -155,6 +186,20 @@
...
 
+  /**
+   * Returns whether the error is the parent of a given error represented by a {@code statusCode}. Parent means that the error
+   * belongs to the same hierarchy or kind of error. For example, 4xx are client errors, 5xx are server errors, etc, so a generic
+   * client error should be the parent of a more specific forbidden one (error code 403).
+   * 
+   * @param statusCode
+   * @return
+   *
+   * @since 1.6.0
```
elrodro83, line 197: `1.5.x`?

src/main/java/org/mule/extension/http/api/error/HttpError.java
szaffarano: This piece of code is the same as before my change.  If `HttpError#valueOf` throws an exception, it's caught silently ("// Do nothing"), and `error` remains null.  In this case, the fallback will search for a category-related error.

src/main/java/org/mule/extension/http/api/error/HttpError.java
szaffarano: Yes, my fault.

src/main/java/org/mule/extension/http/api/error/HttpError.java
szaffarano: You are right, it is a clearer name.

src/main/java/org/mule/extension/http/api/error/HttpError.java
afelisatti: So it would go to `CLIENT_SECURITY`. I think this would work but we will need to be very careful documenting this. People might expect this to capture all client errors.

src/main/java/org/mule/extension/http/api/error/HttpError.java
elrodro83: this is a good opportunity to improve it :)

src/main/java/org/mule/extension/http/api/error/HttpError.java
szaffarano: @afelisatti, you are right.  We've discussed this problem with the team and decided that in order to fix it, we will just add the 5xx errors that were reported.  The more generic solution will be tackled in another JIRA.

src/main/java/org/mule/extension/http/api/error/HttpError.java
szaffarano: done

szaffarano:
--validate

PR #442, opened 2020-05-08
MULE-18396: Use the graceful shutdown timeout to allow persistent connections to do a last request

eze210:


src/test/java/org/mule/test/http/internal/listener/HttpResponseFactoryTestCase.java
```diff
@@ -83,4 +88,52 @@ public void testCursorIteratorCausesInvalidTypeForHttpResponseError() throws Run
     httpResponseBuilder.create(HttpResponse.builder(), new NoInterception(), listenerResponseBuilder, true);
   }
 
+  @Test
```
elrodro83, line 91: add `@Issue` annotation to new tests

src/test/java/org/mule/test/http/internal/listener/HttpResponseFactoryTestCase.java
```diff
@@ -83,4 +88,52 @@ public void testCursorIteratorCausesInvalidTypeForHttpResponseError() throws Run
...
+
+  private void checkForceConnectionCloseHeader(final boolean shouldForceConnectionClose, String defaultValue) {
+    HttpListenerResponseBuilder listenerResponseBuilder = mock(HttpListenerResponseBuilder.class);
+    TypedValue<Object> payload = new TypedValue<>(EXAMPLE_STRING, STRING);
+    when(listenerResponseBuilder.getBody()).thenReturn(payload);
+    when(listenerResponseBuilder.getStatusCode()).thenReturn(OK.getStatusCode());
+
+    MultiMap<String, String> headers = new MultiMap<>();
+    if (defaultValue != null) {
+      headers.put(CONNECTION, defaultValue);
```
elrodro83, line 123: you can use Collections.singletonMap() for this

eze210: But I only want to set the connections header when defaultValue != null; Otherwise the map should be empty

elrodro83: you have `Collections.emptyMap()` for that

eze210: Trying to apply the change, I found it's a MultiMap

PR #446, opened 2020-06-03
MULE-18393: HttpRequester validate entity before retry on remotely cl…

mlischetti:
…osed

### DIFF
--- a/src/main/java/org/mule/extension/http/internal/request/HttpRequester.java
+++ b/src/main/java/org/mule/extension/http/internal/request/HttpRequester.java
@@ -258,7 +258,7 @@ public class HttpRequester {
       } else {
         if (logger.isDebugEnabled()) {
           logger.debug("Sending HTTP message failed with `" + IOException.class.getCanonicalName() + ": " + REMOTELY_CLOSED
-                  + "`. Request will not be retried because entity not support retry.");
+              + "`. Request will not be retried because entity not support retry.");
         }
         shouldRetry = false;
       }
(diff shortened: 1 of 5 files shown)


src/main/java/org/mule/extension/http/internal/request/HttpRequester.java
```diff
@@ -261,6 +268,17 @@ private boolean supportsRetry(String httpMethod) {
     return RETRY_ON_ALL_METHODS || IDEMPOTENT_METHODS.contains(httpMethod);
   }
 
+  private boolean entitySupportRetry(HttpRequest request) {
+    boolean entitySupportRetry = true;
+    if (request.getEntity() != null) {
```
fsgonz, line 273: unify conditions. return true or false according to the condition.

src/main/java/org/mule/extension/http/internal/request/HttpRequester.java
```diff
@@ -247,12 +247,19 @@ private void checkIfRemotelyClosed(Throwable exception, UriParameters uriParamet
...
+  private boolean shouldRetryRemotelyClosed(Throwable exception, int retryCount, HttpRequest httpRequest) {
     boolean shouldRetry = exception instanceof IOException && containsIgnoreCase(exception.getMessage(), REMOTELY_CLOSED)
-        && supportsRetry(httpMethod) && retryCount > 0;
+        && supportsRetry(httpRequest.getMethod()) && retryCount > 0;
     if (shouldRetry) {
-      logger.warn("Sending HTTP message failed with `" + IOException.class.getCanonicalName() + ": " + REMOTELY_CLOSED
-          + "`. Request will be retried " + retryCount + " time(s) before failing.");
+      boolean entitySupportRetry = entitySupportRetry(httpRequest);
+      if (entitySupportRetry) {
+        logger.warn("Sending HTTP message failed with `" + IOException.class.getCanonicalName() + ": " + REMOTELY_CLOSED
```
fsgonz, line 256: I would set this as debug as this was the previous normal behavior. To avoid performance degradation verify that debug is enabled.

PR #449, opened 2020-06-12
HTTPC-124: Apps with listener shouldn't stop servers when they are in other context

eze210:


src/main/java/org/mule/extension/http/internal/listener/HttpListenerProvider.java
```diff
@@ -26,6 +26,7 @@
 import org.mule.runtime.api.scheduler.SchedulerConfig;
 import org.mule.runtime.api.scheduler.SchedulerService;
 import org.mule.runtime.api.tls.TlsContextFactory;
+import org.mule.runtime.core.api.MuleContext;
```
elrodro83, line 29: connectors shouldn't use this

PR #464, opened 2020-08-11

pabloperalta:
Changed the HttpResponseToResult so that the result for a request of the HttpRequester is a repeatable stream because when the payload was read to decide whether to resend or not the request it was being consumed.


pabloperalta:
--validate

pabloperalta:
Will attempt an aproach that would keep the http-connector oblivious of the existence of repeatable streams.

PR #465, opened 2020-08-18

pabloperalta:
After fixing the original error in oauth module the payload was consumed


pabloperalta:
--validate

pabloperalta:
--validate

PR #468, opened 2020-09-01

pabloperalta:
Use repeatable streams when posible in order to prevent the auth mechanism from consuming the payload before it is returned


pabloperalta:
--validate

pabloperalta:
This implementation can be optimized in order to prevent unnecesary performance degradation

PR #482, opened 2020-10-22
HTTPC-128: HTTP Connector sends cookies in separate headers (not RFC6…

eze210:
…265 compliant)

### DIFF
--- a/src/main/java/org/mule/extension/http/internal/request/HttpRequestFactory.java
+++ b/src/main/java/org/mule/extension/http/internal/request/HttpRequestFactory.java
@@ -140,14 +140,7 @@ public class HttpRequestFactory {
         });
 
     if (config.isEnableCookies()) {
-      try {
-        List<String> cookies = config.getCookieManager().get(builder.getUri(), emptyMap()).get(COOKIE);
-        if (cookies != null) {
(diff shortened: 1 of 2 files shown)


src/main/java/org/mule/extension/http/internal/request/HttpRequestFactory.java
```diff
@@ -168,6 +161,27 @@ public HttpRequest create(HttpRequesterConfig config, String uri, String method,
...
+    try {
+      List<String> cookies = config.getCookieManager().get(builder.getUri(), emptyMap()).get(COOKIE);
+      if (cookies != null && cookies.size() > 0) {
+        // The RFC-6265, section 5.4 says:
+        //   If the user agent does attach a Cookie header field to an HTTP
+        //   request, the user agent MUST send the cookie-string (defined below)
+        //   as the value of the header field.
+        //
+        // So we should concatenate the cookies
+        StringBuilder sb = new StringBuilder(cookies.get(0));
```
elrodro83, line 174: try to presize this according to how many headers will be added. The default size is too small and it may grow many times when there are many headers, hurting performance

elrodro83 (changes requested):

elrodro83 (approved):

PR #557, opened 2021-11-11
HTTP-159: Adding connectivity test

eze210:


src/main/java/org/mule/extension/http/api/request/HttpConnectivityValidator.java
```diff
@@ -0,0 +1,201 @@
...
+  @Parameter
+  @Optional(defaultValue = "GET")
+  @Placement(order = 2)
+  private String method;
+
+  /**
+   * The body of the response message
+   */
+  @Parameter
+  @Content(primary = true)
```
marianogonzalez, line 63: content parameters do not apply here

src/main/java/org/mule/extension/http/api/request/HttpConnectivityValidator.java
```diff
@@ -0,0 +1,201 @@
...
+  @Placement(order = 3)
+  private TypedValue<Object> body;
```
marianogonzalez, line 65: user cannot use DW expressions at this point. How is this body supposed to be built?

src/main/java/org/mule/extension/http/api/request/HttpConnectivityValidator.java
```diff
@@ -0,0 +1,201 @@
...
+
+  /**
+   * HTTP headers the message should include.
+   */
+  @Parameter
+  @Optional
+  @Content
+  @NullSafe
+  @Placement(order = 4)
+  private MultiMap<String, String> testHeaders = emptyMultiMap();
```
marianogonzalez, line 75: same here

src/main/java/org/mule/extension/http/internal/request/HttpRequesterProvider.java
```diff
@@ -104,14 +107,32 @@
   @Placement(tab = AUTHENTICATION)
   private HttpRequestAuthentication authentication;
 
+  @Parameter
+  @Optional
+  @Placement(tab = "Test Connection")
+  protected HttpConnectivityValidator testConnection;
```
marianogonzalez, line 113: this should ideally be a `@ParameterGroup(showInDsl = true)`, that is of course unless you have a really good reason to treat this as a pojo

marianogonzalez (changes requested):

src/main/java/org/mule/extension/http/internal/request/HttpRequesterProvider.java
eze210: If it's a pojo it can both be optional and have mandatory parameters, so it's easy to preserve backwards compatibility: if some old application hasn't the test-connection tab, then it doesn't execute the connectivity test and everything works as before.
If it's a pg, we must make all the parameters optional, because otherwise an app which doesn't define the test-connection won't be able to be deployed

src/main/java/org/mule/extension/http/api/request/HttpConnectivityValidator.java
eze210: Removed

src/main/java/org/mule/extension/http/api/request/HttpConnectivityValidator.java
eze210: Now it's a String. It seems to be enough given that it is used for connectivity testing

src/main/java/org/mule/extension/http/api/request/HttpConnectivityValidator.java
eze210: Now it's a list of <String,String> pairs

PR #562, opened 2021-11-25
HTTPC-171: Create HttpRequesterSimpleRequestBuilder for HttpPollingSource and Connectivity Testing

mbuchwald:


src/main/java/org/mule/extension/http/api/request/builder/HttpRequesterSimpleRequestBuilder.java
```diff
@@ -0,0 +1,103 @@
...
+  private Literal<String> pollerBody;
+
+  /**
+   * HTTP headers the message should include.
+   */
+  @Parameter
+  @Optional
+  @NullSafe
+  @DisplayName("Headers")
+  protected MultiMap<String, String> pollerHeaders = emptyMultiMap();
```
eze210, line 44: We should find a prefix that works both for test connection and polling source

eze210: Also, I wonder if this should be a List of some new type instead, as I did in the connectivity test

eze210 (changes requested):

src/main/java/org/mule/extension/http/api/request/builder/HttpRequesterSimpleRequestBuilder.java
```diff
@@ -0,0 +1,104 @@
...
+
+public class HttpRequesterSimpleRequestBuilder {
+
+  /**
+   * The body of the response message
+   */
+  @Parameter
+  @Optional
+  @DisplayName("Body")
+  private Literal<String> requestBody;
```
eze210, line 36: In the connectivity test we're using a `@Text` and here this Literal. Should we remove the body from this class?

src/main/java/org/mule/extension/http/api/request/builder/HttpRequesterSimpleRequestBuilder.java
```diff
@@ -0,0 +1,104 @@
...
+import org.mule.runtime.http.api.domain.message.request.HttpRequestBuilder;
+
+import java.util.Map;
+
+import static java.util.Collections.emptyMap;
+import static java.util.Collections.unmodifiableMap;
+import static org.mule.runtime.api.util.MultiMap.emptyMultiMap;
+import static org.mule.runtime.http.api.server.HttpServerProperties.PRESERVE_HEADER_CASE;
+
+public class HttpRequesterSimpleRequestBuilder {
```
eze210, line 28: Shouldn't we include the path and method too?

mbuchwald: This are not included in the other builder, I would say to remain consistent (also UI-wise)

eze210 (approved):

src/main/java/org/mule/extension/http/api/request/builder/HttpRequesterSimpleRequestBuilder.java
```diff
@@ -0,0 +1,113 @@
...
+
+  /**
+   * Query parameters the request should include.
+   */
+  @Parameter
+  @Optional
+  @DisplayName("Query Parameters")
+  private List<QueryParam> requestQueryParams = emptyList();
+
+  public TypedValue<Object> getRequestBody() {
```
marianogonzalez, line 72: why is this `Object` is the field is actually a String?

src/main/java/org/mule/extension/http/api/request/builder/HttpRequesterSimpleRequestBuilder.java
```diff
@@ -0,0 +1,113 @@
...
+    return TypedValue.of(requestBody);
+  }
+
+  public void setRequestBody(String body) {
```
marianogonzalez, line 76: are the setters really needed? Best practice is for these classes to be immutable (or as close as we can get to that)

src/main/java/org/mule/extension/http/api/request/builder/HttpRequesterSimpleRequestBuilder.java
```diff
@@ -0,0 +1,113 @@
...
+
+  public void setRequestQueryParams(List<QueryParam> queryParams) {
+    this.requestQueryParams = queryParams;
+  }
+
+  public void setRequestUriParams(Map<String, String> uriParams) {
+    this.requestUriParams = uriParams;
+  }
+
+  public HttpRequestBuilder configure(HttpRequesterConfig config) {
```
marianogonzalez, line 108: misleading name. this method gets the impression that `this` instance will get configured when what is actually doing is creating a builder

src/main/java/org/mule/extension/http/api/request/builder/HttpRequesterSimpleRequestBuilder.java
```diff
@@ -0,0 +1,113 @@
...
+
+  public List<SimpleRequestHeader> getRequestHeaders() {
+    return unmodifiableList(requestHeaders);
+  }
+
+  public void setRequestHeaders(List<SimpleRequestHeader> headers) {
+    this.requestHeaders = headers != null ? headers : emptyList();
+  }
+
+  public String replaceUriParams(String path) {
```
marianogonzalez, line 88: confusing name. I would rename to `replaceUriParamsOf`

src/main/java/org/mule/extension/http/internal/request/KeyValuePairUtils.java
```diff
@@ -0,0 +1,29 @@
...
+ * license, a copy of which has been included with this distribution in the
+ * LICENSE.txt file.
+ */
+package org.mule.extension.http.internal.request;
+
+import org.mule.extension.http.api.request.builder.KeyValuePair;
+import org.mule.runtime.api.util.MultiMap;
+
+
+public class KeyValuePairUtils {
```
marianogonzalez, line 13: final

src/main/java/org/mule/extension/http/api/request/builder/KeyValuePair.java
```diff
@@ -0,0 +1,21 @@
...
+
+/**
+ * A single two-strings pair used for key-values used in contexts where using expressions for building maps doesn't
+ * make sense, as within the connection providers.
+ *
+ * @since 1.7
+ */
+public interface KeyValuePair {
+
+  String getKey();
```
marianogonzalez, line 17: if this is API this methods should have jdoc

src/main/java/org/mule/extension/http/api/request/builder/SimpleRequestHeader.java
```diff
@@ -0,0 +1,56 @@
...
+
+import java.util.Objects;
+
+/**
+ * Represents an HTTP Header
+ *
+ * @since 1.7
+ */
+@Alias("requestHeader")
+public class SimpleRequestHeader implements KeyValuePair {
```
marianogonzalez, line 20: why is the class called `SimpleRequestHeader` instead of simply `RequestHeader`?

src/main/java/org/mule/extension/http/internal/request/KeyValuePairUtils.java
```diff
@@ -0,0 +1,29 @@
...
+  }
+
+  /**
+   * Returns a MultiMap representation of an iterable (e.g. a List) of KeyValuePairs
+   * @param iterable
+   * @return
+   */
+  public static MultiMap<String, String> toMultiMap(Iterable<? extends KeyValuePair> iterable) {
+    MultiMap<String, String> asMultiMap = new MultiMap<>();
+    iterable.forEach(pair -> asMultiMap.put(pair.getKey(), pair.getValue()));
```
marianogonzalez, line 26: This is not an issue if this is only used in connectivity testing or polling source, but if this ever makes it into an actual operation that lambda will be a perf problem. I'd replace this for a simple `for`.

marianogonzalez (changes requested):

src/main/java/org/mule/extension/http/api/request/builder/HttpRequesterSimpleRequestBuilder.java
mbuchwald: The problem is potentially doing some unit testing (for the connectivity testing or polling source) without being able to set this manually

marianogonzalez: then I would only add them as protected when needed, specially since this is API

src/main/java/org/mule/extension/http/api/request/builder/SimpleRequestHeader.java
mbuchwald: Because there is already a RequestHeader (which is used somewhere else and we can't reuse here), in the same Package. Having a package for this to then have the same name would be quite misleading

src/main/java/org/mule/extension/http/api/request/builder/QueryParam.java
```diff
@@ -16,18 +16,20 @@
  * 
  * @since 1.1
  */
-public class QueryParam {
+public class QueryParam implements KeyValuePair {
 
   @Parameter
   private String key;
```
marianogonzalez, line 22: Shouldn't you specify that these shouldn't support expressions?

src/main/java/org/mule/extension/http/api/request/builder/QueryParam.java
```diff
@@ -16,18 +16,20 @@
  * 
  * @since 1.1
  */
-public class QueryParam {
+public class QueryParam implements KeyValuePair {
```
marianogonzalez, line 19: all parameter fields should have a jdoc. Keep in mind that jdoc will be used to auto generate documentation so it should be user oriented rather than technical.

marianogonzalez (changes requested):

src/main/java/org/mule/extension/http/api/request/builder/QueryParam.java
mbuchwald: Why? We are currently going to accept strings and that's it, but we would actually like it if the SDK+Studio give support for expressions in here

src/main/java/org/mule/extension/http/api/request/builder/HttpRequesterSimpleRequestBuilder.java
```diff
@@ -0,0 +1,113 @@
...
+  @DisplayName("Headers")
+  protected List<SimpleRequestHeader> requestHeaders = emptyList();
+
+  /**
+   * URI parameters that should be used to create the request.
+   */
+  @Parameter
+  @Optional
+  @DisplayName("URI Parameters")
+  private Map<String, String> requestUriParams = emptyMap();
```
marianogonzalez, line 62: `@NullSafe`

src/main/java/org/mule/extension/http/api/request/builder/HttpRequesterSimpleRequestBuilder.java
```diff
@@ -0,0 +1,113 @@
...
+
+  /**
+   * Query parameters the request should include.
+   */
+  @Parameter
+  @Optional
+  @DisplayName("Query Parameters")
+  private List<QueryParam> requestQueryParams = emptyList();
```
marianogonzalez, line 70: `@NullSafe`

src/main/java/org/mule/extension/http/api/request/builder/QueryParam.java
marianogonzalez: This is going to be used as connectivity testing and sources, both being contexts in which expressions are not available. But most importantly, because this is going to be used for connectivity testing at the connection provider level, an expression here could mislead the SDK to generate a dynamic config unnecessarily, resulting in a performance degradation the user would have a hard time figuring out.

marianogonzalez (changes requested):

src/main/java/org/mule/extension/http/api/request/builder/HttpRequesterTestRequestBuilder.java
```diff
@@ -0,0 +1,113 @@
...
+
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Component that specifies how to create a proper HTTP simple request, that can be used in a connectivity testing.
+ *
+ * @since 1.7
+ */
+public class HttpRequesterTestRequestBuilder {
```
marianogonzalez, line 34: if this is going to be used in a connection provider then it has to define `equals` and `hashCode`

marianogonzalez, line 34: wasn't the point to add `Expression(NOT_SUPPORTED)` on every parameter

marianogonzalez (changes requested):

marianogonzalez (approved):

PR #571, opened 2021-12-09
HTTPC-185: Remove usages of object-to-string-transformer from tests

eze210:


src/test/java/org/mule/test/http/functional/listener/HttpListenerEncodingTestCase.java
```diff
@@ -74,8 +75,16 @@ public void testEncoding() throws Exception {
...
+    assertPayloadIfNotClosed(message, testMessage);
     assertThat(message.getPayload().getDataType().getMediaType().getCharset().get(), is(charset));
   }
 
+  private void assertPayloadIfNotClosed(Message message, String testMessage) throws Exception {
+    try {
+      assertThat(getPayloadAsString(message), is(testMessage));
+    } catch (IllegalStateException e) {
+      // Ignore if closed stream
+    }
```
mbuchwald, line 87: Shouldn't the test fail in this scenario?

mbuchwald (approved):

PR #589, opened 2022-01-06
HTTPC-178 + HTTPC-179: Adding Watermark and Idempotency filtering to HTTP Polling Source

mbuchwald:


src/main/java/org/mule/extension/http/internal/request/HttpPollingSource.java
```diff
@@ -181,53 +179,55 @@ private String getResolvedUri() {
...
+    });
+
+    if (!atLeastOneResult.get()) {
+      LOGGER.debug("Empty result in HTTP Polling Source at {} of uri {}", location.getRootContainerName(), resolvedUri);
+    }
+  }
+
+  private void sendRequest(PollContext<String, HttpResponseAttributes> pollContext) throws InterruptedException {
+    Serializable currentWatermark = pollContext.getWatermark().orElse(null);
+    requestBuilder.updateWatermark(currentWatermark);
```
elrodro83, line 217: i would assume the `requestBuilder` is more of a template... if bothers me that the watermark is being set into it for every request...

src/main/java/org/mule/extension/http/internal/request/HttpPollingSource.java
```diff
@@ -238,22 +238,22 @@ public void poll(PollContext<String, HttpResponseAttributes> pollContext) {
       return;
     }
 
-    sendRequest(pollContext);
+    try {
+      sendRequest(pollContext);
+    } catch (InterruptedException e) {
+      // do nothing
```
elrodro83, line 244: ?

elrodro83 (changes requested):

src/main/java/org/mule/extension/http/internal/request/HttpPollingSource.java
mbuchwald: But the watermark has to be set for every request because it will change

src/main/java/org/mule/extension/http/internal/request/HttpPollingSource.java
```diff
@@ -181,53 +179,55 @@ private String getResolvedUri() {
...
-                   e.getMessage(), e);
+      Result<InputStream, HttpResponseAttributes> result =
+          httpRequester.doSyncRequest(client, config, resolvedUri, method, config.getRequestStreamingMode(),
+                                      config.getSendBodyMode(), config.getFollowRedirects(), client.getDefaultAuthentication(),
+                                      config.getResponseTimeout(), getResponseValidator(), transformationService, requestBuilder,
+                                      true, muleContext, scheduler, injectedHeaders)
+              .get();
+      pollResult(pollContext, result, currentWatermark);
+    } catch (ExecutionException | InterruptedException e) {
+      LOGGER.error("There was an error in HTTP Polling Source at {} of uri '{}'", location.getRootContainerName(), resolvedUri,
```
elrodro83, line 229: the interrupted exeption is not being properly handled here. Ref: https://www.baeldung.com/java-interrupted-exception

elrodro83 (changes requested):

src/main/java/org/mule/extension/http/api/request/builder/SimpleQueryParam.java
```diff
@@ -0,0 +1,78 @@
...
+ * @since 1.7
+ */
+@Alias("requestQueryParam")
+public class SimpleQueryParam implements KeyValuePair {
+
+  /**
+   * Represents the Key of this HTTP Query Parameter
+   */
+  @Parameter
+  private String key;
```
marianogonzalez, line 32: is this used in the context of anything else other than the source? Should this be literal or `@Expression(NOT_SUPPORTED)` ?

src/main/java/org/mule/extension/http/api/request/builder/SimpleRequestHeader.java
```diff
@@ -32,7 +48,15 @@ public String getKey() {
 
   @Override
   public String getValue() {
-    return value;
+    return resolveLiteralExpression(value, expressionLanguage, watermark);
+  }
+
+  public void setExpressionLanguage(ExpressionLanguage expressionLanguage) {
+    this.expressionLanguage = expressionLanguage;
```
marianogonzalez, line 55: let's avoid this kind of things. Injecting API classes such as this with core mule components is not something that would easily fit in new execution models such as voltron. Classes such as this should simply act as DTOs and business logic should be internal to the connector

src/main/java/org/mule/extension/http/internal/request/HttpPollingSource.java
```diff
@@ -87,9 +82,10 @@
 public class HttpPollingSource extends PollingSource<String, HttpResponseAttributes> {
 
   private static final Logger LOGGER = LoggerFactory.getLogger(HttpPollingSource.class);
-  private static final String PAYLOAD_PLACEGHOLDER = "payload";
-  private static final String ITEM_PLACEGHOLDER = "item";
-  private static final String ATTRIBUTES_PLACEGHOLDER = "attributes";
+  public static final String PAYLOAD_PLACEGHOLDER = "payload";
```
marianogonzalez, line 85: `PAYLOAD_PLACEGHOLDER` -> `PAYLOAD_PLACEHOLDER`

src/main/java/org/mule/extension/http/internal/request/HttpPollingSource.java
```diff
@@ -181,53 +180,59 @@ private String getResolvedUri() {
...
-  protected String getId() {
-    return getClass().getSimpleName();
-  }
-
-  private Consumer<PollContext.PollItem<String, HttpResponseAttributes>> getPollingItemConsumer(Result<String, HttpResponseAttributes> item) {
+  private Consumer<PollContext.PollItem<String, HttpResponseAttributes>> getPollingItemConsumer(TypedValue<String> fullResponse,
+                                                                                                Result<TypedValue<?>, HttpResponseAttributes> item,
+                                                                                                Serializable watermark) {
     return pollItem -> {
       LOGGER.debug("Setting Result for {}: {}", location.getRootContainerName(), item.getOutput());
```
marianogonzalez, line 187: shouldn't attributes be logged too?

src/main/java/org/mule/extension/http/internal/request/HttpPollingSource.java
```diff
@@ -181,53 +180,59 @@ private String getResolvedUri() {
...
-      pollItem.setResult(item);
-      // TODO (HTTPC-178 and HTTPC-179 - idempotency and watermarking)
+      pollItem.setResult(toStringResult(item));
```
marianogonzalez, line 188: not following this piece... why `toStringResult`? hasn't the split expression already yielded a String value already?

src/main/java/org/mule/extension/http/internal/request/HttpPollingSourceUtils.java
```diff
@@ -0,0 +1,133 @@
...
+  }
+
+  private static Result<TypedValue<?>, HttpResponseAttributes> toResult(TypedValue<?> item, MediaType mediaType,
+                                                                        HttpResponseAttributes attributes) {
+    return Result.<TypedValue<?>, HttpResponseAttributes>builder().attributes(attributes).output(item).mediaType(mediaType)
+        .build();
+  }
+
+  private static boolean isJavaPayload(MediaType mediaType) {
+    return mediaType.equals(APPLICATION_JAVA.withCharset(mediaType.getCharset().orElse(null)));
```
marianogonzalez, line 59: use `matches` instead of equals. That not only allows you not not worry about the charset, but also about other potential additional properties.

src/main/java/org/mule/extension/http/internal/request/LiteralExpressionUtils.java
```diff
@@ -0,0 +1,55 @@
...
+    String raw = toResolve.getLiteralValue().orElse("");
+    if (!isExpression(raw)) {
+      return raw;
+    }
+    BindingContext context = BindingContext.builder().addBinding(WATERMARK_PLACEGHOLDER, TypedValue.of(watermark)).build();
+    TypedValue<?> result = expressionLanguage.evaluate(raw, context);
+    return result.getValue() != null ? result.getValue().toString() : null;
+  }
+
+  private static boolean isExpression(String value) {
```
marianogonzalez, line 51: doesn't the ExpressionLanguage contains this type of logic already? I think that what'd be best here is to compile all these expressions as part of the source doStart() method. In this way:

1) you don't need to worry about validating the expressions. DW will validate them on its own
2) If an expression is invalid, we can abort the application deployment instead of having a source failing on each poll.

marianogonzalez (changes requested):

src/main/java/org/mule/extension/http/api/request/builder/SimpleQueryParam.java
mbuchwald: It's only on the source

src/main/java/org/mule/extension/http/internal/request/HttpPollingSource.java
mbuchwald: There are a few things here: 
First, the `getItems` returns a stream of results of TypedValue<?> (not string). This is because `split` (from expression manager) returns that type.. and tbh I didn't need to change that (go and change the TypedValue<?> to TypedValue<String>). I use those results also in expressions(e.g. get the id), so I need them as TypedValues. Be it as <?> or <String>. If you think it is better to force it as <String>, no problem. 

But this concrete call is basically to change the result from Result of TypedValue to Result of String, which I need to do regardless of the first paragraph. At some point I tried to leave it as is, but it didn't work well

src/main/java/org/mule/extension/http/api/request/builder/SimpleQueryParam.java
elrodro83: is this intended to be used somewhere other tahn in the source in the future? Otherwise the name is too generic.

src/main/java/org/mule/extension/http/internal/request/HttpPollingSource.java
```diff
@@ -160,74 +169,122 @@ private ResponseValidator getResponseValidator() {
...
+  }
+
   @Override
   protected void doStart() throws MuleException {
     LOGGER.debug("Starting HTTP Polling Source in {}", location.getRootContainerName());
     scheduler = schedulerService.ioScheduler();
     client = clientProvider.connect();
     httpRequester = createHttpRequester(false, muleContext);
-    resolvedUri = getResolvedUri();
+    validateExpressions();
```
elrodro83, line 194: validations should be done at initialise, not start

src/main/java/org/mule/extension/http/internal/request/HttpPollingSource.java
elrodro83: add the explanation as comments in the code

src/main/java/org/mule/extension/http/internal/request/HttpRequester.java
```diff
@@ -135,31 +135,64 @@ public void doRequest(HttpExtensionClient client, HttpRequesterConfig config, St
...
+                       responseValidator, transformationService, requestCreator, checkRetry, muleContext, scheduler,
                        notificationEmitter, streamingHelper, callback,
                        httpRequestFactory.create(config, uri, method, streamingMode, sendBodyMode, transformationService,
-                                                 requestBuilder, authentication, injectedHeaders),
-                       RETRY_ATTEMPTS, injectedHeaders, correlationId);
+                                                 authentication, injectedHeaders, requestCreator),
+                       RETRY_ATTEMPTS, injectedHeaders);
+  }
+
+  public CompletableFuture<Result<InputStream, HttpResponseAttributes>> doSyncRequest(HttpExtensionClient client,
```
elrodro83, line 152: why do you need to return the CompletableFuture instead of the result directly?

src/main/java/org/mule/extension/http/internal/request/HttpRequester.java
```diff
@@ -181,16 +214,21 @@ private void doRequestWithRetry(HttpExtensionClient client, HttpRequesterConfig
...
                 responseValidator.validate((Result) result, httpRequest, streamingHelper);
 
                 Result<Object, HttpResponseAttributes> freshResult = httpResponseToResult
                     .convert(config, muleContext, response, entity, resultInputStreamSupplier, httpRequest.getUri());
 
-                profilingDataProducer
-                    .ifPresent(profilingDataProducer -> profilingDataProducer.triggerProfilingEvent(result, correlationId));
+                profilingDataProducer.ifPresent(
+                                                profilingDataProducer -> profilingDataProducer
+                                                    .triggerProfilingEvent(result,
```
elrodro83, line 227: i think this would be more readable if you extract `correlationId` to a variable

elrodro83 (changes requested):

src/main/java/org/mule/extension/http/api/request/builder/SimpleQueryParam.java
mbuchwald: It is not intended to be used anywhere else... but which name would be better? It's the key of the Query Param, like in QueryParam that was already implemented. If I changed the name, I would still have to show it as "Key" in Studio

src/main/java/org/mule/extension/http/internal/request/HttpPollingSource.java
mbuchwald: But.. how to? 
Sources don't have initialisation phase/method. I can't implement initialisable (SDK doesn't allow source to implement lifecycle interfaces)

elrodro83: good point

elrodro83 (approved):

src/main/java/org/mule/extension/http/api/request/builder/CorrelationData.java
```diff
@@ -0,0 +1,40 @@
+/*
+ * Copyright (c) MuleSoft, Inc.  All rights reserved.  http://www.mulesoft.com
+ * The software in this package is published under the terms of the CPAL v1.0
+ * license, a copy of which has been included with this distribution in the
+ * LICENSE.txt file.
+ */
+package org.mule.extension.http.api.request.builder;
```
marianogonzalez, line 7: why is this api?

src/main/java/org/mule/extension/http/api/request/response/HttpPollingSourceExpressions.java
```diff
@@ -31,8 +31,42 @@
   @DisplayName("Split Expression")
   private Literal<String> splitExpression;
 
+  /**
+   * The expression to retrieve the watermark from the (splitted) response
+   */
+  @Parameter
+  @org.mule.runtime.extension.api.annotation.param.Optional
+  @Text
```
marianogonzalez, line 39: why? doesn't look right.

src/main/java/org/mule/extension/http/api/request/response/HttpPollingSourceExpressions.java
```diff
@@ -31,8 +31,42 @@
...
+  @DisplayName("Watermark Expression")
+  private Literal<String> watermarkExpression;
+
+  /**
+   * The expression to retrieve the id from every (splitted) item
+   */
+  @Parameter
+  @org.mule.runtime.extension.api.annotation.param.Optional
+  @Text
```
marianogonzalez, line 48: why? doesn't look right.

src/main/java/org/mule/extension/http/api/request/response/HttpPollingSourceExpressions.java
```diff
@@ -31,8 +31,42 @@
   @DisplayName("Split Expression")
   private Literal<String> splitExpression;
```
marianogonzalez, line 32: `@Expression(REQUIRED)`?

src/main/java/org/mule/extension/http/api/request/response/HttpPollingSourceExpressions.java
```diff
@@ -31,8 +31,42 @@
...
 
+  /**
+   * The expression to retrieve the watermark from the (splitted) response
+   */
+  @Parameter
+  @org.mule.runtime.extension.api.annotation.param.Optional
+  @Text
+  @DisplayName("Watermark Expression")
+  private Literal<String> watermarkExpression;
```
marianogonzalez, line 41: `@Expression(REQUIRED)`?

src/main/java/org/mule/extension/http/api/request/response/HttpPollingSourceExpressions.java
```diff
@@ -31,8 +31,42 @@
...
+
+  /**
+   * The expression to retrieve the id from every (splitted) item
+   */
+  @Parameter
+  @org.mule.runtime.extension.api.annotation.param.Optional
+  @Text
+  @DisplayName("Id Expression")
+  private Literal<String> idExpression;
```
marianogonzalez, line 50: `@Expression(REQUIRED)`?

marianogonzalez (changes requested):

marianogonzalez (approved):

PR #602, opened 2022-02-08
[w-10660947] composer connectivity testing

zzcontr:
Added new method HttpConnectivityValidator for use by composer http wrapper.

### DIFF
--- a/src/main/java/org/mule/extension/http/api/request/HttpConnectivityValidator.java
+++ b/src/main/java/org/mule/extension/http/api/request/HttpConnectivityValidator.java
@@ -34,6 +34,7 @@ import org.mule.runtime.extension.api.annotation.param.ParameterGroup;
 import org.mule.runtime.extension.api.annotation.param.display.DisplayName;
 import org.mule.runtime.extension.api.annotation.param.display.Placement;
 import org.mule.runtime.extension.api.runtime.operation.Result;
+import org.mule.runtime.http.api.HttpConstants;
 import org.mule.runtime.http.api.client.auth.HttpAuthentication;
 import org.mule.runtime.http.api.domain.message.request.HttpRequest;
 import org.mule.runtime.http.api.domain.message.response.HttpResponse;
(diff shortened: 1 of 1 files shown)


src/main/java/org/mule/extension/http/api/request/HttpConnectivityValidator.java
```diff
@@ -130,6 +131,24 @@ private void validateResult(HttpRequest request, Result result) {
     getResponseValidator().validate(result, request);
   }
 
+  //This method is used for composer purposes only
+  public void validate(HttpExtensionClient client, HttpConstants.Protocol protocol, String host, Integer port)
```
eze210, line 135: Why isn't the other validate() method enough?

eze210, line 135: This method repeats code from the other methods in this class

src/main/java/org/mule/extension/http/api/request/HttpConnectivityValidator.java
```diff
@@ -130,6 +131,24 @@ private void validateResult(HttpRequest request, Result result) {
...
+      throws ExecutionException, InterruptedException, ResponseValidatorTypedException {
+    String pathWithUriParams = replaceUriParams(requestPath, requestBuilder.getRequestUriParams());
+    String uriString = protocol.getScheme() + "://" + host + ":" + port + pathWithUriParams;
+    responseValidator = new SuccessStatusCodeValidator("200..299");
```
eze210, line 139: Here it's replacing the old responseValidator. It would change the behaviour of this validator for future validate() calls, and that doesn't sound like something that validate() should do

src/main/java/org/mule/extension/http/api/request/HttpConnectivityValidator.java
```diff
@@ -130,6 +131,24 @@ private void validateResult(HttpRequest request, Result result) {
     getResponseValidator().validate(result, request);
   }
 
+  //This method is used for composer purposes only
```
eze210, line 134: Which are those purposes?

eze210 (changes requested):
Missing test cases

src/main/java/org/mule/extension/http/api/request/HttpConnectivityValidator.java
zzcontr: removed it

eze210 (approved):

PR #603, opened 2022-02-17
W-10667956: Validate and document TLSv1.3 usage

pabloperalta:


src/test/java/org/mule/test/http/functional/tls/HttpTlsContext13TestCase.java
```diff
@@ -0,0 +1,67 @@
...
+
+  @Override
+  protected String getConfigFile() {
+    return "http-tls-1.3-test.xml";
+  }
+
+  @BeforeClass
+  public static void assumeJdkIsCompatible() {
+    String[] javaVersion = System.getProperty("java.version").split("_");
+    Assume.assumeTrue(SystemUtils.isJavaVersionAtLeast(JavaVersion.JAVA_1_8)
```
asanguinetti, line 59: static imports

src/test/resources/http-tls-1.3-test.xml
```diff
@@ -0,0 +1,76 @@
...
+                               keyPassword="123456" password="123456"/>
+            </tls:context>
+        </http:request-connection>
+    </http:request-config>
+    <flow name="tls-1.3-client-external" doc:id="e82d891f-3f22-4d51-ba87-4b9bdd140c6b">
+        <http:listener doc:name="Listener" config-ref="HTTP_Listener_config_unsecure" path="/client-external"/>
+        <http:request method="GET" doc:name="Akamai" config-ref="HTTP_Request_configuration"
+                      url="https://tls13.akamai.io/"/>
+        <logger level="INFO" doc:name="Logger" message="#[payload]"/>
+    </flow>
```
asanguinetti, line 64: We are not using this flow and its associated configs. I understand we don't want to use it for automated test to avoid depending on connections to external resources. Maybe your intention was to leave it for manual troubleshooting. In that case, shouldn't we at least comment it out and leave an explanation on how and when to enable it for debugging? I'm concerned it could be misleading as to what we are testing if we leave it like this, what do you think?

pabloperalta: I was just checking if the test would pass with the pipeline's JDK :P Sorry! Now that it passed I will clean everything up

pabloperalta:
I left them for reference. Maybe someone wants to reuse them in the future. I like the Idea of adding a script.

asanguinetti (dismissed):

asanguinetti (approved):

PR #623, opened 2022-05-27
W-11214346: Fix HttpListenerConfigFunctionalTestCase

eze210:


src/test/java/org/mule/test/http/functional/listener/HttpListenerConfigFunctionalTestCase.java
```diff
@@ -155,20 +157,36 @@ private HttpResponse callAndAssertStatus(String url, HttpStatus expectedStatus)
...
+        throw new RuntimeException("Could not find network interface different from localhost");
+      } else {
+        return preferSiteLocal(candidates);
+      }
     } catch (SocketException e) {
       throw new RuntimeException(e);
     }
   }
 
+  private static String preferSiteLocal(List<? extends InetAddress> candidates) {
```
fsgonz, line 181: what is the non site-local address that also appears as candidate?

eze210: It's the IP from "this" host in the Zscaler virtual network. If you try to ping that IP, it times out:
```
$ ping 100.64.0.1
PING 100.64.0.1 (100.64.0.1): 56 data bytes
Request timeout for icmp_seq 0
Request timeout for icmp_seq 1
Request timeout for icmp_seq 2
Request timeout for icmp_seq 3
Request timeout for icmp_seq 4
^C
--- 100.64.0.1 ping statistics ---
6 packets transmitted, 0 packets received, 100.0% packet loss
```

fsgonz: 👍

fsgonz (approved):

PR #666, opened 2022-12-23
useless change

eze210:


eze210:
This PR was aimed to test which checks were activated

PR #668, opened 2023-01-06
W-12176357: Change Span Kind according to OTel semantic conventions in HTTP Connector

sofiamorseletto:


sofiamorseletto:
--validate

PR #673, opened 2023-02-16
Add Headers as Attributes for Otel traces

panizzag:
GUS Associated: [W-12558102](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00001LPLeGYAX/view)

This commit adds the possibility that the headers processed by the HTTP listener and by the HTTP requester are added as attributes of the spans initialized during the Otel implementation. This new feature takes into account the option of omitting headers that you do not want to register by using a property that is configured at the config element level of both components. An example of use is as follows:

- HTTP Listener

`<http:listener-config name="HTTP_Listener_config" doc:name="HTTP Listener config" doc:id="f83efe45-479a-498f-adc1-8848f33ed380" skipHeadersOnTracing="client_id, client_secret">`

- HTTP Requester

`<http:request-config name="HTTP_Request_configuration" doc:name="HTTP Request configuration" doc:id="3a18fe34-e53b-4ca4-80e2-229cb386f1bb" preserveHeadersCase="true" skipHeadersOnTracing="client_id, client_secret">`

- Example cUrl request
`curl --location --request GET 'localhost:8081/test' \
--header 'X-Example-Header: Test' \
--header 'client_id: thisIsAnExampleClientID' \
--header 'client_secret: thisIsAnExampleClientSecret'`

The final result is the following:

- **Before the change** - Attributes registered for the example call
   - Listener:
   image
   
   - Requester:
   image

- **After the change** - Attributes registered for the example call
   - Listener
     image

   - Requester 
     image

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -15,7 +15,7 @@
     <groupId>org.mule.connectors</groupId>
     <artifactId>mule-http-connector</artifactId>
     <packaging>mule-extension</packaging>
-    <version>1.8.0-SNAPSHOT</version>
+    <version>1.8.1-SNAPSHOT</version>
 
     <name>HTTP Connector</name>
     <description>A Mule extension that provides HTTP server and client functionality</description>
--- a/src/main/java/org/mule/extension/http/api/listener/server/HttpListenerConfig.java
+++ b/src/main/java/org/mule/extension/http/api/listener/server/HttpListenerConfig.java
@@ -9,6 +9,7 @@ package org.mule.extension.http.api.listener.server;
 import static java.util.Optional.empty;
 import static java.util.Optional.of;
 import static org.mule.runtime.api.meta.ExpressionSupport.NOT_SUPPORTED;
+import static org.mule.runtime.api.meta.ExpressionSupport.SUPPORTED;
 import static org.mule.runtime.api.util.Preconditions.checkArgument;
 
 import org.mule.extension.http.api.listener.headers.HttpHeadersException;
@@ -66,12 +67,25 @@ public class HttpListenerConfig implements Initialisable {
   @Expression(NOT_SUPPORTED)
   private boolean rejectInvalidTransferEncoding;
 
+  /**
+   * W-12558102
+   * Indicates what headers should not be exported as attributes when generating Open Telemetry traces. By default, common headers associated with credentials are skipped.
+   *
+   * @since 1.8.0
+   */
+  @Parameter
+  @Optional(defaultValue = "client_id, client_secret, Authorization")
+  @Expression(SUPPORTED)
+  private String skipHeadersOnTracing;
+
   private HttpHeadersValidator httpHeaderValidators;
 
   @Override
   public void initialise() throws InitialisationException {
     basePath = sanitizePathWithStartSlash(this.basePath);
     httpHeaderValidators = new InvalidTransferEncodingValidator(rejectInvalidTransferEncoding);
+    //W-12558102
+    skipHeadersOnTracing = this.skipHeadersOnTracing;
   }
 
   public ListenerPath getFullListenerPath(String listenerPath) {
@@ -98,5 +112,10 @@ public class HttpListenerConfig implements Initialisable {
    */
   public void validateHeaders(MultiMap<String, String> headers) throws HttpHeadersException {
     httpHeaderValidators.validateHeaders(headers);
+  }
+
+  //W-12558102
+  public String getSkipHeadersOnTracing() {
+    return skipHeadersOnTracing;
   }
 }
--- a/src/main/java/org/mule/extension/http/internal/listener/profiling/tracing/HttpListenerCurrentSpanCustomizer.java
+++ b/src/main/java/org/mule/extension/http/internal/listener/profiling/tracing/HttpListenerCurrentSpanCustomizer.java
@@ -15,9 +15,14 @@ import static org.slf4j.LoggerFactory.getLogger;
 import org.mule.extension.http.api.HttpRequestAttributes;
 import org.mule.extension.http.internal.request.profiling.tracing.HttpCurrentSpanCustomizer;
 import org.mule.sdk.api.runtime.source.DistributedTraceContextManager;
+import org.mule.runtime.api.util.MultiMap;
+import com.google.common.collect.ArrayListMultimap;
+import com.google.common.collect.ListMultimap;
 
 import java.net.URI;
 import java.net.URISyntaxException;
+import java.util.List;
+import java.util.Arrays;
 
 import org.slf4j.Logger;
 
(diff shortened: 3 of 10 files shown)


pom.xml
```diff
@@ -15,7 +15,7 @@
     <groupId>org.mule.connectors</groupId>
     <artifactId>mule-http-connector</artifactId>
     <packaging>mule-extension</packaging>
-    <version>1.8.0-SNAPSHOT</version>
+    <version>1.8.1-SNAPSHOT</version>
```
fsgonz, line 18: We change the snapshot version once we release a connector. Let's change it again to 1.8.0

src/main/java/org/mule/extension/http/api/listener/server/HttpListenerConfig.java
```diff
@@ -66,12 +67,25 @@ public class HttpListenerConfig implements Initialisable {
...
+  private String skipHeadersOnTracing;
+
   private HttpHeadersValidator httpHeaderValidators;
 
   @Override
   public void initialise() throws InitialisationException {
     basePath = sanitizePathWithStartSlash(this.basePath);
     httpHeaderValidators = new InvalidTransferEncodingValidator(rejectInvalidTransferEncoding);
+    //W-12558102
+    skipHeadersOnTracing = this.skipHeadersOnTracing;
```
fsgonz, line 88: Is this line correct?

src/main/java/org/mule/extension/http/internal/listener/profiling/tracing/HttpListenerCurrentSpanCustomizer.java
```diff
@@ -15,9 +15,14 @@
 import org.mule.extension.http.api.HttpRequestAttributes;
 import org.mule.extension.http.internal.request.profiling.tracing.HttpCurrentSpanCustomizer;
 import org.mule.sdk.api.runtime.source.DistributedTraceContextManager;
+import org.mule.runtime.api.util.MultiMap;
```
fsgonz, line 18: We have an order for the imports:

static mule packages
static java, javax packages
static other packages

after that:

mule packages
java, javax packages
rest of the packages.

src/main/java/org/mule/extension/http/internal/listener/profiling/tracing/HttpListenerCurrentSpanCustomizer.java
```diff
@@ -41,16 +46,24 @@ public class HttpListenerCurrentSpanCustomizer extends HttpCurrentSpanCustomizer
   private final String host;
   private final int port;
 
-  private HttpListenerCurrentSpanCustomizer(HttpRequestAttributes attributes, String host, int port) {
+  //W-12558102
```
fsgonz, line 49: I would remove the comment for the feature.

src/main/java/org/mule/extension/http/internal/listener/profiling/tracing/HttpListenerCurrentSpanCustomizer.java
```diff
@@ -41,16 +46,24 @@ public class HttpListenerCurrentSpanCustomizer extends HttpCurrentSpanCustomizer
...
+  private final List<String> skipAttributesList;
+
+  private HttpListenerCurrentSpanCustomizer(HttpRequestAttributes attributes, String host, int port, String skipAttributes) {
     this.attributes = attributes;
     this.host = host;
     this.port = port;
+    //W-12558102
+    this.skipAttributesList = Arrays.asList(skipAttributes.split(",", -1));
```
fsgonz, line 57: use a static import (whenever possible)

fsgonz (changes requested):

fsgonz:
Thanks very much for the PR, @panizzag! 
I have added some cosmetic request for changes for the moment (just in case you have time to add them) 
Can you verify if you some unit tests for that. We will be working on adding some integration tests.
Probably we will have to discuss a little bit the new feature with the team but it clear and straightforward.

src/main/java/org/mule/extension/http/internal/listener/profiling/tracing/HttpListenerCurrentSpanCustomizer.java
```diff
@@ -110,4 +121,20 @@ protected String getSpanKind() {
...
     return attributes.getListenerPath();
   }
+
+  /**
+   * W-12558102
+   * This provides a transparent way to obtain the definitive list of headers to add in the spans associated with Otel tracing.
+   * This list will have all headers except those that have been skipped via the skipHeadersOnTracing property.
+   */
+  @Override
+  public MultiMap<String, String> getHeaders() {
```
sofiamorseletto, line 131: This method seems to be the same for the HTTP Listener and the requester, could its definition be moved to HttpCurrentSpanCustomizer?

pom.xml
panizzag: Done

src/main/java/org/mule/extension/http/api/listener/server/HttpListenerConfig.java
panizzag: No, it's not. Removed.

src/main/java/org/mule/extension/http/internal/listener/profiling/tracing/HttpListenerCurrentSpanCustomizer.java
panizzag: Thank you. I applied it for other classes as well.

src/main/java/org/mule/extension/http/internal/listener/profiling/tracing/HttpListenerCurrentSpanCustomizer.java
panizzag: Done. Should I remove all the other similar comments?

src/main/java/org/mule/extension/http/internal/listener/profiling/tracing/HttpListenerCurrentSpanCustomizer.java
panizzag: Done

panizzag:
> 
Thanks, @fsgonz for the review and suggestions. I have expanded the existing test case to include the new features. I have also fixed the test as it was failing before applying the changes. I'd appreciate it if the team can give it a new look as there might be coding guidelines and styles that I'm not used to. Thanks

fsgonz:
Thanks very much. We'll also have to verify if we follow the semantics conventions 
https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md#http-request-and-response-headers
And if it is enabled by default. Thanks again

PR #675, opened 2023-03-23
W-12337859: Change Span Status according to OTel semantic conventions in HTTP Connector

sofiamorseletto:


sofiamorseletto:
--validate

fsgonz:
--validate

PR #824, opened 2024-04-03
Some changes to un-ignore and/or fix some tests after the reactor-netty client impl of the service

jesica-fera:
This PR shouldn't be merged before: https://github.com/mulesoft-labs/mule-http2-service/pull/16
This is just against test/netty branch were we keep the changes for the tests related to the new HTTP Netty Service


functional-tests/http/.java-version
```diff
@@ -0,0 +1 @@
+1.8
```
eze210, line 1: This file addition doesn't sound intentional

eze210 (changes requested):

functional-tests/http/.java-version
jesica-fera: Oops! Sorry, removed

eze210 (approved):

PR #830, opened 2024-04-30
W-15400264: Modify some tests to handle new slight behaviour changes of Netty

jesica-fera:
Some of the tests were accordingly modified to support change of messages or other small stuff that started happening with Netty

### DIFF
--- a/functional-tests/http/src/test/java/org/mule/test/http/functional/requester/HttpRequestMaxConnectionsTestCase.java
+++ b/functional-tests/http/src/test/java/org/mule/test/http/functional/requester/HttpRequestMaxConnectionsTestCase.java
@@ -24,10 +24,8 @@ import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 
 import org.eclipse.jetty.server.Request;
-import org.junit.Ignore;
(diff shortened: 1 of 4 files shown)


functional-tests/http/src/test/java/org/mule/test/http/functional/requester/HttpRequestMaxConnectionsTestCase.java
```diff
@@ -58,6 +57,7 @@ public void maxConnections() throws Exception {
   private Thread processAsynchronously(final Flow flow) {
     Thread thread = new Thread(() -> {
       try {
+        System.out.println("Sent one request");
```
eze210, line 60: Remove the print or replace it with a logger

eze210 (changes requested):

functional-tests/http/src/test/java/org/mule/test/http/functional/requester/HttpRequestMaxConnectionsTestCase.java
jesica-fera: It's not needed, removed

eze210 (approved):

PR #932, opened 2024-10-30
W-16606364 Add proxy and ntlm coverage for streaming 

aronee2008:
Tests added - 

- NTLM proxy with streaming
- Global Anonymous Proxy with streaming 
- Inner Anonymous Proxy with streaming 
- Global Username/Password Proxy with streaming 
- Inner Username/Password Proxy with streaming 
- Validate streaming using client and server with proxy


functional-tests/http/src/test/resources/http-streaming-request-proxy-config.xml
```diff
@@ -0,0 +1,66 @@
...
+                <http:proxy host="localhost" port="${proxyPort}"/>
+            </http:proxy-config>
+        </http:request-connection>
+    </http:request-config>
+
+    <flow name="InnerAnonymousProxy">
+        <http:listener config-ref="listenerConfig" path="/test/InnerAnonymousProxy" />
+        <http:request config-ref="configInnerAnonymousProxy" path="requestPath"/>
+    </flow>
+
```
hyerans, line 38: extra blank

functional-tests/http/src/test/resources/http-streaming-request-proxy-config.xml
```diff
@@ -0,0 +1,66 @@
...
+    <!-- Inner Username/Password Proxy -->
+    <http:request-config name="configInnerUserPassProxy" basePath="basePath">
+        <http:request-connection host="localhost" port="${httpPort}">
+            <http:proxy-config>
+                <http:proxy host="localhost" port="${proxyPort}" username="theUsername" password="thePassword" />
+            </http:proxy-config>
+        </http:request-connection>
+
```
hyerans, line 58: extra blank

functional-tests/http/src/test/java/org/mule/test/http/functional/requester/streaming/HttpStreamingRequestNtlmProxyConfigTestCase.java
```diff
@@ -0,0 +1,70 @@
...
+import org.apache.http.client.fluent.Request;
+import org.apache.http.client.fluent.Response;
+import org.apache.http.entity.StringEntity;
+import org.junit.Before;
+import org.junit.BeforeClass;
+import org.junit.Rule;
+
+
+@Stories({@Story(PROXY), @Story(NTLM), @Story(STREAMING)})
+public class HttpStreamingRequestNtlmProxyConfigTestCase
```
eze210, line 41: This test caught my attention because it's testing the proxy authentication in a strange way... I see that you copied it from an older test, so it's not your fault, but I will ask you to add a comment here so we know what is it doing in the future. The comment should be something like:
The server that is working as a "proxy" (and executing the authentication) here is the one we usually use as the "target", but it works for the test purposes.

eze210 (changes requested):

eze210 (approved):

PR #960, opened 2024-11-29
W-17333982: fix for HTTP:TIMEOUT error

neetya26:
Screenshot 2024-12-02 at 6 26 51 PM
Screenshot 2024-12-03 at 12 01 24 PM

Attaching before and after images for the fix


sf-himmatbir (approved):

eze210 (changes requested):
Please add the corresponding MUnit test case

PR #961, opened 2024-11-29
W-17333982: fix for HTTP:TIMEOUT error

neetya26:
Screenshot 2024-12-02 at 6 26 51 PM
Screenshot 2024-12-03 at 12 01 24 PM

Attaching before and after images for the fix

### DIFF
--- a/mule-http-connector/src/main/java/org/mule/extension/http/internal/request/HttpRequester.java
+++ b/mule-http-connector/src/main/java/org/mule/extension/http/internal/request/HttpRequester.java
@@ -14,7 +14,7 @@ import static java.util.Optional.empty;
 import static java.util.Optional.ofNullable;
 import static org.apache.commons.lang3.StringUtils.containsIgnoreCase;
 import static org.mule.extension.http.api.error.HttpError.CONNECTIVITY;
-import static org.mule.extension.http.api.error.HttpError.TIMEOUT;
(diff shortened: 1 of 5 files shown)


mule-http-connector/src/main/java/org/mule/extension/http/internal/request/HttpRequester.java
```diff
@@ -262,7 +262,7 @@ private void doRequestWithRetry(HttpExtensionClient client, HttpRequesterConfig
             }
 
             logger.error(getErrorMessage(httpRequest));
-            HttpError error = exception instanceof TimeoutException ? TIMEOUT : CONNECTIVITY;
+            HttpError error = exception instanceof TimeoutException ? REQUEST_TIMEOUT : CONNECTIVITY;
```
sf-himmatbir, line 265: As discussed, Currently we are throwing HTTP:Timeout in case of no response, after this change we will throw HTTP:REQUEST_TIMEOUT. 
As per earlier implementation customers might be expecting HTTP:Timeout here so this might break few things on their side. It would suggest to keep it same.
For the scenario where status code is 408, we can try if HTTP:Timeout can be thrown there as well to make sure it is same everywhere.
If above is not possible then can go ahead with REQUEST_TIMEOUT at both places and need to update the documentation as well.

sf-himmatbir (approved):

eze210 (changes requested):
Please add the corresponding MUnit test case

PR #969, opened 2024-12-13
W-17370109: Add tests for the situation where requester finds an error after committing the response payload

eze210:


eze210:
This test was added in integration, because it tests a feature of the service

PR #1008, opened 2025-03-04
W-17769292 Unignore passing munit tests for cookie handling and redirects for netty impl

aronee2008:
Passes all munit tests for http netty service for cookies and redirect cases 
image
image

3/18
munit tests with netty:
image
functional tests with netty: 
image


mule-http-connector/src/test/munit/cookies/request-cookies-redirect-test-case.xml
```diff
@@ -26,9 +26,7 @@
         <http:request-connection host="localhost" port="${dynamic.port}"/>
     </http:request-config>
 
-    <!-- TODO (W-17405565): Un-ignore this test for Netty impl -->
-    <munit:test name="cookiesWithMultipleRedirect" description="Cookies are replaced during the redirect mechanism"
-                ignore="#[p('mule.http.service.implementation') == 'NETTY']">
+    <munit:test name="cookiesWithMultipleRedirect" description="Cookies are replaced during the redirect mechanism]" ignore="#[Munit::muleVersionPriorTo('4.9.3')]">
```
eze210, line 29: Only ignore the test for versions older than 4.9.3 when the implementation is NETTY. Otherwise, you're ignoring this test for GRIZZLY implementation too.

eze210: And do the same for the rest of the tests

eze210 (changes requested):

eze210 (approved):

PR #1024, opened 2025-04-03
W-17430471 Move 100 continue connector tests to munit 

aronee2008:


hyerans:
Shouldn't we remove this to-do //TODO W-17430471 Migrate 100 continue test to Munit from [HttpRequestExpect100ContinueHeaderTestCase](https://github.com/mulesoft/mule-http-connector/blob/8ce2b46d4bfaf81bf1d9aeea8d9f2c0328da6a74/functional-tests/http/src/test/java/org/mule/test/http/functional/requester/HttpRequestExpect100ContinueHeaderTestCase.java#L27)?

hyerans:
Also, The original test (`HttpRequestExpect100ContinueHeaderTestCase`) covered three main client-side scenarios. It looks like the `testExpect100Continue` test case in the client-side munit test file covers the first scenario (`requestFlow`). However, I don't see tests covering the other two original scenarios:
- the case using `requestFlowNoHeaders` (successful request without explicit Expect header),
- the timeout case using `requestFlowWithLowTimeout`.
Separately, the `testExpect100ContinueFor417` test seems to introduce new coverage for malformed headers, rather than directly replacing one of the missing original scenarios. Please let me know if I missed anything or if these scenarios are covered elsewhere.

hyerans:
The original tests were parameterized to run against both persistent and non-persistent connections. Are we addressing this in the MUnit test as well?

aronee2008:
> * requestFlowNoHeaders

I have added the coverage missing, `requestFlowNoHeaders` scenario does not belong to the `expect-continue` scenario as it only tests the vanilla http-request scenario.

